PR #11, opened 2016-01-18
Use Path.Combine for cross-platform certificate path

martijnboland:
With this fix, unit tests also pass on OS X when running 'dnx test'.


leastprivilege:
thanks!

PR #57, opened 2016-03-15
Convert the JSON ScopesString into seperate claims

smallplanet-it:
I noticed a strange, in my opinion wrong behavior in the TokenValidator class.
Let’s say you have requested a valid access_token with more than one scope, for example „openid“, „profile“, „email“ and „api1“

The convert of the JSON ScopesString into seperate claims is not correctly implemented. As it is at the moment it works only if you have an access_token with only one „openid“ scope.


leastprivilege:
Hi thanks.

But you changed to things in one PR. We'd also need a unit test that shows that your change is working.

leastprivilege:
I guess this is not relevant anymore.

PR #101, opened 2016-05-23
.NET Core RC2 - .NET 4.6 Support

PawelGerr:
Please add .NET 4.6 support. These change requires the changes from pull request https://github.com/IdentityModel/IdentityModelv2/pull/3


leastprivilege:
i updated all the packages - give it a go and see if that works better for you (and open a new issue if you encounter problems)

thanks!

PR #108, opened 2016-05-27
Added PublicOrigin back to IdentityServerOptions

ericgreenmix:
Added PublicOrigin back to IdentityServerOptions and modified the BaseUrlMiddleware to set the request context host to use this if present. This enables proper origin domains when sitting behind a proxy.

### DIFF
--- a/src/IdentityServer4/Configuration/Options/IdentityServerOptions.cs
+++ b/src/IdentityServer4/Configuration/Options/IdentityServerOptions.cs
@@ -38,6 +38,14 @@ namespace IdentityServer4.Core.Configuration
         public string SiteName { get; set; }
 
         /// <summary>
+        /// Gets or sets the public origin for IdentityServer (e.g. "https://yourserver:1234").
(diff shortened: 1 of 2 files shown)


ericgreenmix:
CLA signed.

brockallen:
Is there a reason this should be in IdSvr and not done at the hosting layer? This was our feeling and why we did not carry that over. But maybe we overlooked something.

ericgreenmix:
Thanks for the quickly reply Brock. The more that I think about it, you guys are right. I decided to work around this by adding in middleware before IdSvr. Going to close this PR.

```
                app.Use(async (httpcontext, next) =>
                { 
                    var origin = new Uri(Configuration["PUBLIC_ORIGIN"]);
                    httpcontext.Request.Scheme = origin.Scheme;
                    httpcontext.Request.Host = new HostString(origin.Authority);
                    await next();
                });
                app.UseIdentityServer();
```

PR #152, opened 2016-07-05
Bug Fix

PivotalAnimal:
requestedClaimTypes was being passed in non-null as an empty collection. This was causing RequestedClaimTypes to be set which ultimately ended in a missing 'sub' claim error message in the browser.

### DIFF
--- a/src/IdentityServer4/Models/Contexts/ProfileDataRequestContext.cs
+++ b/src/IdentityServer4/Models/Contexts/ProfileDataRequestContext.cs
@@ -81,7 +81,7 @@ namespace IdentityServer4.Models
             Client = client;
             Caller = caller;
 
-            if (requestedClaimTypes == null)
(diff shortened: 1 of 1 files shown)


PivotalAnimal:
I fixed a small problem in my project that was causing an empty collection to be passed into this method. However, I still think this is probably a bug. Else more checking for errant conditions.

leastprivilege:
thanks. Does the above commit work for you?

PivotalAnimal:
Sure. I think it is safe to assume that an empty collection indicates _AllClaimsRequested_ be set to true. It is sort of up to the Scopes / Claims provider to make certain that they have a valid configuration. Further downstream, I don't know the full ramifications. Is there anything that would impact security? On the other-hand, allowing an empty collection to be set to _RequestedClaimTypes_ and detecting a missing sub claim from the client-side via an error also seems legit. That is how I found the problem to begin with.

leastprivilege:
ok. thanks

PR #168, opened 2016-07-21
Support for public clients

leastprivilege:
Public clients don't need a client secret to request tokens from the token endpoint


leastprivilege:
@brockallen please review

brockallen:
:shipit:

Issue #185, opened 2016-07-27
Figure out cross references in docs


leastprivilege:
https://github.com/IdentityServer/IdentityServer4/pull/184

mpoettgen:
Could it be that avoiding underscore (_) within the label/hyperlink target could do the trick? it looks like it has a special meaning, so could mean trouble.

http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#hyperlink-targets

PR #229, opened 2016-08-23
Ref links fix

### LINKED ISSUE #182: broken link in docs
https://identityserver4.readthedocs.io/en/dev/start/mvc.html

 use _this_ sample; link is http://127.0.0.1:8000/start/clients.html#defining-a-server-side-web-application-e-g-mvc-for-use-authentication-and-delegated-api-access

NetChris:
I think this fixes #182 and #185.  See my build of the docs here, with the references working:

http://netchrisidentityserver4.readthedocs.io/en/dev/start/mvc.html

I had to remove both the internal underscores _and_ the dashes.  Those didn't seem to work either.  I ended up making the reference name "startClientsMVC" instead of "start-clients-mvc" (as suggested in [the docs](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#hyperlink-targets)).

(HT to @mpoettgen on the "underscores are bad" idea)

Sorry for the multiple commits here.  This is the nature of the beast when you need to change/commit/build to see how the docs come out on the other end.

### DIFF
--- a/docs/start/clients.rst
+++ b/docs/start/clients.rst
@@ -77,7 +77,7 @@ This client uses the so called implicit flow to request an identity and access t
         },
     };
 
-.. _start_clients_mvc:
+.. _start-clients-mvc:
 Defining a server-side web application (e.g. MVC) for use authentication and delegated API access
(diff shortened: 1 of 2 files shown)


leastprivilege:
thanks!

Issue #212, opened 2016-08-19
Is Token.Client needed?


cbeall:
The Token model currently contains a Client property. I'd like to propose that property be exchanged for just a ClientId string (which already partially exists as a read only property which calls the Client property).

Currently, the Client property isn't read anywhere in the codebase except for accessing the ClientId or the AccesssTokenType. The AccessTokenType already exists in the Token model (albeit as a string and not the AccessTokenType enum).

The reason I'd like to see it removed is that it complicates the storage of token. If the client information and tokens are persisted through different means (for example a more typical database for clients and a reliable cache for tokens), it leaves one of three options when retrieving a token.

1. Make a call to both stores and attach client information to token.
2. When storing the token, copy all of the client information within the token. 
3. Only include the ClientId string within the token and trust the codebase won't change later by asking for an additional field not being stored.

Any of which are doable, but if there is no need for that for Token to have all of the Client data, its seems preferable to removed Client and replaced with ClientId. I'd be happy to pitch in if that helps the cause.

leastprivilege:
that's a good point - we will discuss this for the RC

leastprivilege:
We reworked persistence - it should be a lot easier now. Give it a try. Feel free to open a new issue.

PR #233, opened 2016-08-24
Retrieval of Clients and Scopes for Token and AuthorizationCode

cbeall:
The following is in reference to #212. 

I really like what @brockallen did with the persistence. By handling the serialization of the various types and making them as small as possible, it makes the job of implementing the store much easier. 

I have a suggestion that I hope compliments this effort. @leastprivilege said to open a new issue to follow up, but I hope a PR is alright.

The PersistentGrantSerializer is current using a JsonConverter to make calls to ClientStore and ScopeStore when needed. I think there are a few issues with this.
1. Not all use cases require the full client or scope objects making some of these calls unnecessary.
2. Making async calls in a non async JsonConverter method doesn't feel very clean.
3. The ScopeConverter only has the ability to lookup one scope at a time, and potentially makes multiple calls when one would do.

To mitigate this, I first modified properties in Token and AuthorizationCode.

* In Token, removed Client and replaced it with ClientId and AccessTokenType.
* In AuthorizationCode, removed Client and replaced it with ClientId and Lifetime (Lifetime already exists in Token and RefreshToken)

I then identified cases downstream that need the full Client or Scope and retrieve them there. I move the calls to the ClientStore and/or ScopeStore there.  Those cases were 

* When TokenResponseGenerator processes an authorization code and needs to return an IdentityToken
* When TokenResponseGenerator creates an access code for a flow with an authorization code 
* When TokenValidator validates a reference token.

To me, its cleaner to get the full client/scope only in the cases when they are needed. This avoids potentially unneeded calls to the store. It can also be done in a way without the use of the AsyncHelper class and instead uses a "async all the way" approach.

These changes all fall under the coverage of existing tests. There was no need to modify any tests beyond adding the new properties during the test setup. 

Please let me know what you think.


brockallen:
Thanks a lot Craig. This is definitely an improvement.

cbeall:
You're welcome... I'm happy to have been able help.

PR #241, opened 2016-08-29
Making scope parameter optional at the token endpoint 

leastprivilege:
(for client credentials, password and extension grants)

* if allowed scopes is configured, client can omit scope parameter
* if only `AllowAccessToAllScopes` is configured, client must specify scopes explicitly


patrikulus:
After update to version 1.3.1 it stopped working for password grants :(

brockallen:
> After update to version 1.3.1 it stopped working for password grants :(

please open a new issue if you think this is a bug.

PR #255, opened 2016-09-02
Move semicolon to fix syntax error

PeterOrneholm:
The code in the example for "Defining a client for server to server communication" did not compile.


leastprivilege:
thanks!

PR #261, opened 2016-09-05

crussell85:
I think this paragraph should be after adding the identity server middleware. If you try to visit the "identity" endpoint before this the "Authorize" tag causes a .NET error. 

An unhandled exception was thrown by the application.
System.InvalidOperationException: No authentication handler is configured to handle the scheme: Automatic

Once the identityserver middleware the 401 response is received as expected


brockallen:
Yep, good point. Thanks!

PR #286, opened 2016-09-12
Scope claim is always a string array

leastprivilege:


brockallen:
:shipit: :+1:

adminnz:
There appears to be another issues.

When going through the tutorial in the documentation if you do the switching to hybrid flow (https://identityserver4.readthedocs.io/en/dev/quickstarts/5_hybrid_and_api_access.html) and have a user with multiple roles (e.g. "Admin" and "Developer").

They get combined into 1 claim (type = "role")  with an array value of `["Admin",  "Developer"]`
Where previously when doing Implicit GrantType they would come across as two claims one with "Admin" and the other with "Developer" both claims having the type "role".

brockallen:
Ok, we can look into it. // @leastprivilege

brockallen:
@adminnz Looks like it's a Microsoft bug: https://github.com/aspnet/Security/issues/976

PR #320, opened 2016-09-18
add IdP as part of authorize request validation

### LINKED ISSUE #319: Consider ignoring IdP acr_value if IdP not in client allowed list
This would remove the pressure from the UI to have this logic. We simply won't pass it along in the sign in message.

brockallen:
fixes #319

// @leastprivilege


leastprivilege:
thanks for your support!

PR #321, opened 2016-09-18
use client config on login page (enable local; allowed providers)

### LINKED ISSUE #318: Add logic/support to bypass HRD on login screen

brockallen:
Incorporate client settings on login page for enable local logins, and external provider restriction list. This required re-working how the login view model is constructed.

Fixes #318

// @leastprivilege


leastprivilege:
another great contribution!

PR #325, opened 2016-09-20
Port "private_key_jwt" from IdentityServer3

### LINKED ISSUE #27: Port JWT client authentication
...needs new Microsoft.IdentityModel (RC2)

uncleyo:
Port "private_key_jwt" from IdentityServer3

- Full certificate X5c header support removed from the implementation for
simplicity (since they are extremely rarely used and not well supported
anyway).
- Improve SecretParser: Keep trying other parsers if found secret is "NoSecret"
- Add integration tests for client assertion

Closes #27


leastprivilege:
Cool. thanks! I will have a closer look over the weekend!

PR #362, opened 2016-10-01
Fixed the Configure method contents

MartinZikmund:
Fixed the documentation quick start example - previously there was no call to `app.UseIdentityServer` which is necessary to make the example actually work (`.well-known` discovery endpoint)

### DIFF
--- a/docs/quickstarts/1_client_credentials.rst
+++ b/docs/quickstarts/1_client_credentials.rst
@@ -130,6 +130,7 @@ It must be added **before** MVC, e.g.::
         loggerFactory.AddConsole(Configuration.GetSection("Logging"));
         loggerFactory.AddDebug();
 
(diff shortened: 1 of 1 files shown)


docs/quickstarts/1_client_credentials.rst
```diff
@@ -130,6 +130,7 @@ It must be added **before** MVC, e.g.::
         loggerFactory.AddConsole(Configuration.GetSection("Logging"));
         loggerFactory.AddDebug();
 
+        app.UseIdentityServer();
```
leastprivilege: That does not make sense. You are adding identityserver to the API pipeline.

MartinZikmund: Oh my... I got the sample mixed up. Sorry about that :-) .

Issue #19, opened 2016-01-23
Multiple IdSrvs mapped on one website


edwinf:
I am starting to port our code to v4 to see if we run into any issues so we can provide feedback / questions per [this comment](https://github.com/IdentityServer/IdentityServer4/issues/7#issuecomment-171646779)

Using IdSrv 3 we have 2 identity servers mapped to the same host (`host.com/users`, `host.com/clients` - the main reasons for this are completely different user and view services between the two).  After reviewing the sample, pouring over documentation and looking at the v4 codebase itself, I cannot figure out how to set up the same type of configuration in v4.  I'm assuming I'm missing something simple, but in case I'm not, figured I'd ask if this type of configuration is still possible?

brockallen:
This is a fundamental design flaw of ASP.NET Core's DI system -- there is no official way to have distinct DI systems for different branches in the pipeline. We (and many others) have provided this feedback to Microsoft. They say they're aware of it and will be doing something at some point to allow for this. BTW, they call this "multi-tenancy" support, which is just... an awful.. awful name.

joeaudette:
I am also facing this issue. I had a similar issue with asp.net identity, but the solution I used for that doesn't seem possible with IdentityServer4.

The difference is that asp.net identity uses IOptions to get options used when [setting the cookie options and auth scheme from IApplicationBuilder](https://github.com/aspnet/Identity/blob/dev/src/Microsoft.AspNetCore.Identity/BuilderExtensions.cs) and they use 

    IOptions<IdentityOptions>

which allowed me to implement a custom implementation of the same so that when the actual options are accessed from the IOptions.Value I could resolve the correct settings based on the request url. even though the IOptions itself is a singleton the Value when accessed can return a different result based on  the request.

The way IdentityServerOptions is a singleton and not wrapped in IOptions poses a challenge.

I'm using [sasskit](https://github.com/saaskit/saaskit) for multi-tenancy which enables me to support tenants resolved from the first folder segment of the url (as one possibility), so in startup I wire up the per folder tenant authentication like this:

            app.UseMultitenancy<cloudscribe.Core.Models.SiteSettings>();

            app.UsePerTenant<cloudscribe.Core.Models.SiteSettings>((ctx, builder) =>
            {
                builder.UseCloudscribeCoreDefaultAuthentication(
                    loggerFactory,
                    multiTenantOptions,
                    ctx.Tenant);
                
            });

[...]

PR #382, opened 2016-10-07
potentially solves #19

joeaudette:
to support custom matching with an additional url segment for
multi-tenant


brockallen:
Given that this PR is so old, I think it makes sense to close it and follow up on the open issue so we can understand the real ask here.

PR #384, opened 2016-10-07
Rename InMemoryLoginService.cs

patriksvensson:
While browsing the source code I noticed that the class `InMemoryUserLoginService` isn't located in `InMemoryUserLoginService.cs` as I expected, but in `InMemoryLoginService.cs`.

If this for any reason is by design, please ignore this PR.


leastprivilege:
thanks!

PR #386, opened 2016-10-08
AuthenticationScheme and SignInScheme is case sensitive

patriksvensson:
I might be "holding it wrong", but I can't get this to work properly unless spelling `Cookies` with a capital `C`.


patriksvensson:
Nevermind this. I was completely wrong...

Issue #253, opened 2016-09-01
Add caching for introspection


leastprivilege:

brockallen:
Looks like it might be best for this on the `IPersistedGrantService` for the `GetReferenceTokenAsync`.

brockallen:
Also, caching on IsActive?

brockallen:
>  Also, caching on IsActive?

I decided that I'd not want a built in IsActive cache -- it's simple enough for a customer to do that themselves.

brockallen:
The other place we could do caching is on the `TokenValidator.ValidateAccessToken` which is called from both introspection and userinfo endpoints. It's an easy cache key (the token itself).

brockallen:
Postpone -- we need something more clever and into the token validator itself.

brockallen:
Too scenario specific to provide a default implementation that will work for most scenarios. Can be done manually with a custom IReferenceToken store that does caching.

PR #387, opened 2016-10-08
add token validation cache

brockallen:
Helps with #253


brockallen:
BTW if this is used, we need to rework logging again since the inner token validator won't be called and thus we won't get our informational on certain endpoints (like introspection and user info).

Issue #394, opened 2016-10-09
Discuss IPersistedGrantService


brockallen:
Should this really be 5 different interfaces (notice the 5 groupings of APIs)?

https://github.com/IdentityServer/IdentityServer4/blob/dev/src/IdentityServer4/Services/IPersistedGrantService.cs

Of course there's overlapping functionality, but a base class would solve that. The reason I mention it is that to implement things like this: https://github.com/IdentityServer/IdentityServer4/issues/379, it's be much more convenient to have just a few APIs to have to override/implement via a decorator.

The underlying persisted grant store would be unaffected.

brockallen:
I started to do this work and was not happy with it. In short, this approach ends up being like IdSvr3 with the 4 different store interfaces. 

What if instead we removed all these APIs on the IPersistedGrantService, and instead made them all extension methods to the IPersistedGrantStore instead?

brockallen:
done

Issue #379, opened 2016-10-07
Consider "ConsentCookie" ?


leastprivilege:
In the default UI - similar to 2FA "Trust this browser"

brockallen:
I think the best/easiest place to implement this would be as a custom `IPersistedGrantService` with `StoreUserConsentAsync` and `GetUserConsentAsync` checking the DB as well as a cookie.

brockallen:
Yes, it's now possible.

PR #428, opened 2016-10-20
reworked persisted grant service/store as extension methods

brockallen:
#394 // @leastprivilege


brockallen:
crap, this PR doesn't help with #379

PR #432, opened 2016-10-21
generate EF Migrations to be more aligned with the file new project

DerAlbertCom:
In an newly created AspNet.Core Project with AspNetIdentity the Migrations are in the folder Data\Migrations
These changes align the IdentityServerMigrations with the file new project expierence.


brockallen:
ok, thanks.

PR #433, opened 2016-10-21
Fixed Claims property on InMemoryUser initializer

ChrisSimmons:
At least in an ASP.NET Core app, the code as presented doesn't compile.  Added `new []` to ensure that the `InMemoryUser.Claims` property is a valid `IEnumerable<Claim>`.


leastprivilege:
thanks!

PR #434, opened 2016-10-21
Added clearing of JWT handler claim type mappings

ChrisSimmons:
Unless I'm mistaken and missed something in a previous slide, this is needed to produce the screenshot further down which shows the "sub" claim (as opposed to http://schemas.xmlsoap.org/ws/2005/05/identity/claims/nameidentifier).


ChrisSimmons:
I can't decipher what went wrong with the Travis build.  There's something vague about "An error occurred while generating the build script."  I don't know how modifying documentation would break a build.

leastprivilege:
thanks!

PR #445, opened 2016-10-27
Removed exception from XML doc

ChrisSimmons:
The method doesn't actually throw System.NotImplementedException.  I think this might be something either GhostDoc or Exceptional (ReSharper plug-in) added when the method was originally stubbed out.


brockallen:
thanks

PR #457, opened 2016-11-02
DefaultSessionIdService.GetCookieName is now configurable

khelben:
from IdentityServerOptions, the default is "idsvr.session" to not break existing configurations


brockallen:
i broke this in my recent update, but it's a good idea. i'll merge it in when i can.

brockallen:
looks like this got added in already. thanks!

PR #482, opened 2016-11-16
Updated docs for code that doesn't break the program

alesten:
I found that if I tried adding clients before scope my project wouldn't run


brockallen:
Given the recent PR, this is now unnecessary; Closing. Thanks.

PR #486, opened 2016-11-17
add clarity on AddInMemoryUsers method

MarcusPozzan:
the flow of the tutorials and quickstarts, do not explain the in memory usage of inmemoryuser. however the method is clearly defined in the actual quickstart for EF.
referecing the method and location in the quickstart.
open to complete rewrite, but the method was not discussed earlier or in the aspnetcore identity quickstart.


brockallen:
I don't see how this helps. The quickstarts assume the progression, so you should have seen the InMemoryUsers API in the previous quickstarts.

PR #502, opened 2016-11-24
Fixes #487 - Name of the session cookie is derived from EffeciveAuthenticationScheme

### LINKED ISSUE #487: Make name of the session tracking cookie configurable
I am trying to put together a multi-tenant instance of identity server using idsvr4. A key requirement for me is to be able to issue cookies with tenant-specific names. I have managed to change the cookie named `idsvr4` to a custom name just by changing `AuthenticationScheme` and registering my own cookie middleware. However, this does not change the name of the cookie issued for session tracking as this name is hardcoded (https://github.com/IdentityServer/IdentityServer4/blob/55ea89002f64079893637724c334b3e00d2862b2/src/IdentityServer4/Services/Default/DefaultSessionIdService.cs#L63)

There is a comment that says this needs to be read from config. When is this planned to be fixed? And can this be changed so that cookie name is derived from the one specified in the cookie middleware?

schatekar:
Deriving name of the session cookie from `EffectiveAuthenitcationScheme` makes it configurable. 

I have run all the unit tests locally and they pass. I do not know how to run integration tests locally. Appreciate if someone can help with that.

### DIFF
--- a/src/IdentityServer4/Services/Default/DefaultSessionIdService.cs
+++ b/src/IdentityServer4/Services/Default/DefaultSessionIdService.cs
@@ -5,7 +5,9 @@
 using IdentityModel;
 using Microsoft.AspNetCore.Http;
 using System.Threading.Tasks;
+using IdentityServer4.Configuration;
 using Microsoft.AspNetCore.Http.Features.Authentication;
+using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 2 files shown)


schatekar:
I think I know what's causing this to fail. On it now.

brockallen:
Thanks!

Issue #498, opened 2016-11-24
Update ServiceCollection/Builder docs


leastprivilege:

leastprivilege:
started with general structure

Issue #496, opened 2016-11-22
Consider removing ClaimsRule


brockallen:

brockallen:
done on config branch

Issue #495, opened 2016-11-22
Consider removing IncludeAllClaimsForUser


brockallen:

brockallen:
done on config branch

Issue #468, opened 2016-11-07
Code name: resource configuration model


brockallen:
:wink: :wink:

brockallen:
@leastprivilege first working prototype of config rework on branch: https://github.com/IdentityServer/IdentityServer4/tree/config

brockallen:
done on config branch

PR #509, opened 2016-11-29
new resource based configuration changes

brockallen:
#468
#498
#496
#495


leastprivilege:
epic!

FelschR:
what's the reason why IncludeAllClaimsForUser was removed?

brockallen:
It allowed people to be sloppy. If you want the same semantics, it's easy enough to do that logic in a custom `IProfileService`.

PR #552, opened 2016-12-07
Improve definition of standard scopes from OIDC spec.

diogodamiani:
Defining static read-only properties for standard scopes from OIDC spec. No need to instantiate (use the new operator) each resource.


leastprivilege:
Dude - thanks. But how about you start a discussion first before you change 16 files?

leastprivilege:
..and why is that an improvement exactly?

diogodamiani:
I'm sorry I did not start a discussion before. :(

I think it's clearer for anyone who uses the code not to instantiate an object that is defined in another class (IdentityResources). When you look for standard types, they are usually in the same class that defines the type.

brockallen:
The reason we changed them to classes to create is to make it easier if the display name or description needed to be changed.

diogodamiani:
Do you mean "needed to be changed" by you or by the consumers of the API? The way it's in the PR allow changes on the default objects if it's necessary, by the spec (you), but not by consumers. 

Anyway, as Dominick said, I should have started a discussion before have sent the PR.

brockallen:
i think we're going to stick with what we have. thanks anyway.

PR #563, opened 2016-12-09
docs: update ef quickstart image to rc5

timscriv:
Image for Entity framework database was out of date with new resources instead of scopes.


brockallen:
ah, thanks!

PR #684, opened 2017-01-12
Always invoke profile service in DefaultClaimsService

### LINKED ISSUE #683: Always invoke profile service in DefaultClaimsService

akamud:
- Users can now control which claims are added to id_token and access_token, regardless of scope claims configured

Closes #683


brockallen:
we need to think about if we put this on a 1.1 branch, or if we put it into 1.0.2

brockallen:
we'll add this in 1.0.2 -- we took what you provided as a starting point. thanks.

PR #709, opened 2017-01-18
Update 2_resource_owner_passwords.rst

swinterrowd:
The TestUser class will not be recognized by the compiler unless the using statement is added to the file first.


brockallen:
thanks

PR #713, opened 2017-01-18
Update 2_resource_owner_passwords.rst

swinterrowd:
In the following text from this documentation:
"When you send the token to the identity API endpoint, you will notice one small
but important difference compared to the client credentials grant. The access token will
now contain a ``sub`` claim which uniquely identifies the user.

The presence (or absence) of the ``sub`` claim let's the API distinguish between calls on behalf
of clients and calls on behalf of users."
The statements are unclear as to how to identify the "sub" claim. It would help to tell the reader how the "sub" claim can be seen and verified. I see that the access token returned for the ResourceOwnerPassword request now has more characters in the returned access token for the ClientCredential request. Are the extra characters in the access token the "sub" claim described. I am not sure about where the "sub" claim is.


brockallen:
thanks!

Issue #722, opened 2017-01-20
Where is bower.json in Host project?


ateregulov:
My gitignore contains "/wwwroot/lib" path.

totht91:
Yes it's missing. The /wwwroot/lib gitignore is a common usa-case.

ateregulov:
I created pull request #723

brockallen:
We don't have one, as our gitignore does not filter that path. If you want to use bower, then add it in your project.

PR #723, opened 2017-01-20
add bower.json

ateregulov:
For issue #722


brockallen:
Why?

ateregulov:
It is front-end development guidelines.

totht91:
It's like nuget, if you don't commit the packages folder, you sholdn't commit the wwwroot/libs too.

brockallen:
The host is a sample/testing project, not a production project. I don't see the real need for bower here. If you want bower in your systems, then by all means use it.

brockallen:
Thanks for this, but as I said I think this is something you would just do in your own hosting project.

PR #748, opened 2017-01-26
Improve grammar

PaluMacil:
This colloquial abbreviation is folksy and should not appear in serious writing. Additionally, without the word "are" before "gonna" the grammar is also incorrect.


brockallen:
thanks

PR #761, opened 2017-02-01
Update extension_grants.rst

rmja:
TokenErrors -> TokenRequestErrors

### DIFF
--- a/docs/topics/extension_grants.rst
+++ b/docs/topics/extension_grants.rst
@@ -78,14 +78,14 @@ It's the job of the extension grant validator to handle that request by validati
 
             if (string.IsNullOrEmpty(userToken))
             {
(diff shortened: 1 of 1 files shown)


leastprivilege:
thanks!

rmja:
@leastprivilege Should there be some handling in the case where the sub claim is null? For example if the token was issued in a client credential flow. In fact, how is that even meant to work?

I was thinking along the lines of:

```
var subject = result.Claims.FirstOrDefault(c => c.Type == JwtClaimTypes.Subject)?.Value;

            if (subject != null)
            {
                context.Result = new GrantValidationResult(subject, "delegation");
            }
            else
            {
                // what to put here?
            }
```

leastprivilege:
wrong place for discussion. open an issue. thanks

rmja:
Sorry, Opened #775.

Issue #768, opened 2017-02-02
JsonWebKey not allowed as SigningCredentials


igarciaolaizola:
``JsonWebKey`` cannot be used as `SigningCredentials`. An exception will be thrown saying that the signing key needs to be asymmetric, even if the Key included in the JsonWebKey is asymmetric.
````
JsonWebKey jsonWebKey = JsonWebKey.Create(json);
SigningCredentials credential = new SigningCredentials(jsonWebKey, jsonWebKey.Alg);
services
    .AddIdentityServer().
    .AddSigningCredential(credential)
...
````
The reason of this is that `JsonWebKey` does not inherate from `AsymmetricSecurityKey` and therefore, this validation here fails: [see Crypto.cs](https://github.com/IdentityServer/IdentityServer4/blob/e46dde73e3a47976ee40f8a9912cdbf705e50067/src/IdentityServer4/Configuration/DependencyInjection/BuilderExtensions/Crypto.cs#L27)
````
if (!(credential.Key is AsymmetricSecurityKey))
{
    throw new InvalidOperationException("Signing key is not asymmetric");
}
````

**Comment by leastprivilege** [comment]
Then create an asymmetric key from the jwks?

**Comment by igarciaolaizola** [comment]
Yes, that could be a possible solution. But since `System.IdentityModel.Tokens.JwtJwtSecurityTokenHandler `can deal with JsonWebKeys for signing and validation and it already implements the extraction of an RSA/Elliptic Curve key from the JsonWebKey, that solution makes me feel like I am reinventing the wheel.

A better solution might be change how the validation for asymmetric keys is done. Something like:
```
if (!(credential.Key is AsymmetricSecurityKey 
    || (credential.Key is JsonWebKey && ((JsonWebKey)credential.Key).HasPrivateKey)))
{
    throw new InvalidOperationException("Signing key is not asymmetric");
}
```

Meanwhile I could add an specific extension for `JsonWebKey` in order to skip this validation:
```
public static IIdentityServerBuilder AddSigningCredential(this IIdentityServerBuilder builder, JsonWebKey jsonWebKey)
{
    if (jsonWebKey == null) throw new ArgumentNullException(nameof(jsonWebKey));

    if (!jsonWebKey.HasPrivateKey)
    {
        throw new InvalidOperationException("JSON Web Key does not have a private key.");
    }

    SigningCredentials credential = new SigningCredentials(jsonWebKey, jsonWebKey.Alg);
[...]

PR #772, opened 2017-02-02
allow JsonWebKeys that contain an asymmetric key

igarciaolaizola:
Allow a JsonWebKey as SigningCredentials when it contains an asymmetric key.
As proposed in #768


leastprivilege:
Could you add a test please?

igarciaolaizola:
> Could you add a test please?

I've added 2 test cases, JWK with asymmetric key (successful case) and JWK with symmetric key (failure case).

leastprivilege:
thanks!

PR #777, opened 2017-02-03
Add option for having additional claims included when issuing a clien…

rmja:
…t jwt from tools

### DIFF
--- a/src/IdentityServer4/Extensions/IdentityServerToolsExtensions.cs
+++ b/src/IdentityServer4/Extensions/IdentityServerToolsExtensions.cs
@@ -12,7 +12,7 @@ namespace IdentityServer4
 {
     public static class IdentityServerToolsExtensions
     {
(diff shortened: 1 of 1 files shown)


leastprivilege:
You see that this calls

return await tools.IssueJwtAsync(lifetime, claims);

in turn?

rmja:
Yes I saw that. I just thought that it would be nice to include it in the convenience extension.

leastprivilege:
IMHO the current API surface enables all the scenarios.

Thanks anyways!

rmja:
Fine with me, I am just spreading my thoughts as I progress in implementing my setup😵 

As an alternative, could we then expose the httpcontext as a public property on the tools instead of it being an internal field? The reason being that I need it in my own custom tools extension method. (I am currently getting it via reflection.)

leastprivilege:
ouch ;)

To get to HttpContext - either pass it in - or inject `IHttpContextAccessor`

PR #825, opened 2017-02-16
Validate scoped services

user1336:
Hi,

So in ASP.NET Core 1.1 it is possible to do the following:
```
public IServiceProvider ConfigureServices(IServiceCollection services)
{
    return services.BuildServiceProvider(validateScopes: true);
}
```
Setting this parameter to true will validate the correct lifetime of the dependencies.
This gave a few errors in `TestService()` in combination with the IdentityServer4.EF package. 
The services are being resolved from the root ServiceProvider which means they will be Singleton.

Kinda related: 
https://github.com/IdentityServer/IdentityServer4.EntityFramework/pull/57


giggio:
You should dispose the `serviceProvider` at the end.

leastprivilege:
thanks!

Issue #629, opened 2016-12-22
Make grant result more flexible


leastprivilege:
* add additional claims on the fly
* change scopes
* change token type

...

leastprivilege:
"no subject" scenarios - add client claims on dynamically. see #775

leastprivilege:
also #599

Make token type settable?!

leastprivilege:
..and lifetime

leastprivilege:
done

PR #831, opened 2017-02-19
Add more flexibility to token request pipeline [WIP]

leastprivilege:
see #629

Sometimes it would be useful to be able to change certain parameters on the fly during a token request (e.g. from a custom validator or an extension grant), e.g.

* scopes
* token lifetime
* token type
* client claims


leastprivilege:
@brockallen The idea is to hang off an "overrides" object from the `ValidatedTokenRequest` which can be populated by a grant validator or a custom token request validator. If something is set there, this will be honoured by the token creation plumbing.

Not super happy about the word `Overrides` - but it's a start ;)

leastprivilege:
back to the drawing board ;)

PR #874, opened 2017-03-04
Change client allowed grant types to ICollection

brockallen:
For consideration and open to discussion. This is breaking, so we'd want to merge it into 2.0.


brockallen:
FYI: `GrantTypes.List` removed in favor of `ICollection<string>` for `Client.AllowedGrantTypes`.

PR #876, opened 2017-03-04
fix misspell cacue to cache

siyo-wang:
In Docs/Confgiuration/startup.rst have typo, the cache misspell to cacue.


leastprivilege:
thanks!

PR #887, opened 2017-03-07
Pass LoginUrl/LogoutUrl from IdentityServerOptions to auth cookie

cwe1ss:
When an unauthenticated user tries to access an authenticated page from an IdentityServer app (e.g. for changing a password) the user gets redirected to "/Account/Login?ReturnUrl=...". This is the default value in the ASP.NET Core cookie authentication middleware and in IdentityServer (although it's lowercase "/account/login" there) so everything works with the default settings. 

Unfortunately, changes to `identityServerOptions.Authentication.LoginUrl`, `identityServerOptions.Authentication.LogoutUrl` and `identityServerOptions.Authentication.LoginReturnUrlParameter` are currently not passed to the cookie authentication middleware. This results in the user still being redirected to "/Account/Login". My workaround for now is to use my own cookies in Startup.cs.

This PR would change that by passing the option values to the cookie middleware.

### DIFF
--- a/src/IdentityServer4/Hosting/CookieMiddleware.cs
+++ b/src/IdentityServer4/Hosting/CookieMiddleware.cs
@@ -30,6 +30,9 @@ namespace IdentityServer4.Hosting
                         SlidingExpiration = options.Authentication.CookieSlidingExpiration,
                         ExpireTimeSpan = options.Authentication.CookieLifetime,
                         CookieName = IdentityServerConstants.DefaultCookieAuthenticationScheme,
+                        LoginPath = options.UserInteraction.LoginUrl,
+                        LogoutPath = options.UserInteraction.LogoutUrl,
+                        ReturnUrlParameter = options.UserInteraction.LoginReturnUrlParameter
                     });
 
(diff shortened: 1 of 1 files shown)


src/IdentityServer4/Hosting/CookieMiddleware.cs
```diff
@@ -30,6 +30,9 @@ public static void ConfigureCookies(this IApplicationBuilder app)
                         SlidingExpiration = options.Authentication.CookieSlidingExpiration,
                         ExpireTimeSpan = options.Authentication.CookieLifetime,
                         CookieName = IdentityServerConstants.DefaultCookieAuthenticationScheme,
+                        LoginPath = options.UserInteraction.LoginUrl,
```
brockallen: you'd want something like this:
```
if (url.IsLocalUrl())
{
    if (url.StartsWith("~/")) url = url.Substring(1);
}
```

and then assign the url to the `LoginPath` and/or `LogoutPath`. If they're not local, then don't assign them.

cwe1ss: I didn't know that you could use external urls for that. I updated the code.

brockallen:
Thanks!

PR #908, opened 2017-03-11
Update videos.rst

janskoruba:
Adding the link with new Pluralsight course where is presented the configuration of the IdentityServer4.


leastprivilege:
thanks. but this belongs under training. I'll add it manually

PR #915, opened 2017-03-14
Changed the comment to the corresponding default.

user1336:
Changed the comment to the corresponding property value.


leastprivilege:
thanks!

Issue #931, opened 2017-03-16
Invalid 'X-Frame-Options' header encountered : 'ALLOWFROM is not a recognized directive


jp7677:
Hi there,

When performing a logout I'm seeing the following error in Chrome:
`Invalid 'X-Frame-Options' header encountered when loading 'http://localhost:3060/connect/endsession/callback?[some params] : 'ALLOWFROM http://localhost:3060' is not a recognized directive. The header will be ignored.`

As far as I know the error is correct because it should be ALLOW-FROM instead of ALLOWFROM, see
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options

Looking at your code this looks like a typo in https://github.com/IdentityServer/IdentityServer4/blob/e46dde73e3a47976ee40f8a9912cdbf705e50067/src/IdentityServer4/Endpoints/Results/EndSessionCallbackResult.cs#L119

Could you eventually rename ALLOWFROM to ALLOW-FROM EndSessionCallbackResult,
see https://github.com/IdentityServer/IdentityServer4/pull/932

Regards and thanks for your great project here,
Jens

brockallen:
Merged, thanks!

PR #932, opened 2017-03-16
Fix typo in the X-Frame-Options directive.

jp7677:
It should be ALLOW-FROM instead of ALLOWFROM. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options

This fixes https://github.com/IdentityServer/IdentityServer4/issues/931


brockallen:
Yep, a typo. Thanks!

PR #933, opened 2017-03-16
Add nuget package restore note to apis.rst

jeremycook:


leastprivilege:
Thanks. But I don't think this belongs here. This applies to any nuget package you ever gonna add.

PR #940, opened 2017-03-17
Note to add IdentityModel to MVC client

jeremycook:


leastprivilege:
thanks - but outdated now ;)

PR #957, opened 2017-03-21
Add option to persists temp key

leastprivilege:
saves the temp key in file


stuartleeks:
Looking at this issue it looks like this was added to v1.4. I've just updated to v1.5 but don't see the `AddTemporarySigningCredential` overload with the `persist` parameter. Am I missing something?

Thanks!

leastprivilege:
Use `AddDeveloperSigningCredential`

PR #963, opened 2017-03-21
Add validated request to token result

brockallen:
@leastprivilege thoughts on doing this? it's convenient for events :)


leastprivilege:
OK

PR #981, opened 2017-03-25
Added a note to the external authentication docs.

serpent5:
ASP.NET Core Identity combined with External Authentication requires the SignInScheme to be set differently.


serpent5:
It might be better to take an IOptions<IdentityOptions> into Configure and sniff out the value of Cookies.ExternalCookieAuthenticationScheme, but I didn't want to overcomplicate the note.

leastprivilege:
thanks!

PR #1030, opened 2017-04-06
Create new collection when implying all scopes if none requested

### LINKED ISSUE #1026: AllowedScopes grows on cached Client objects with no requested scope and offline access enabled
### Issue
When using the client store caching on clients where allow offline access is enabled, and making token requests without supplying a scope, the `offline_access` scope is added to the list of allowed scopes. The problem is here in `TokenRequestValidator.ValidateRequestedScopes()`:

```
var clientAllowedScopes = _validatedRequest.Client.AllowedScopes;
if (!clientAllowedScopes.IsNullOrEmpty())
{
    if (_validatedRequest.Client.AllowOfflineAccess)
    {
        clientAllowedScopes.Add(IdentityServerConstants.StandardScopes.OfflineAccess);
    }
    scopes = clientAllowedScopes.ToSpaceSeparatedString();
    _logger.LogTrace("Defaulting to: {scopes}", scopes);
}
else
{
    LogError("No allowed scopes configured for {clientId}", _validatedRequest.Client.ClientId);
    return false;
}
```

Since the `clientAllowedScopes` is just a reference to the collection returned by `Client.AllowedScopes` and the Client object is cached, eventually the  generated scope string exceeds the default length limitation, resulting in an `invalid_scope` response.

### Steps to reproduce the problem

1. Register IdentityServer using `AddInMemoryCache()` and `AddClientStoreCache()`
2. Make multiple token requests with password, client credentials, or extension grant types.
3. Watch the `Clients.AllowedScopes` grow until it exceeds the length limit.

chrisowhite:
Fixes #1026. When using the provided in-memory client store, or enabling
in-memory client store cache, this prevents the offline_access scope
from being continually added to the Client.AllowedScopes collection when
AllowOfflineAccess = true and no scopes were requested for resource
owner, client credentials, and extension grant types.

### DIFF
--- a/src/IdentityServer4/Validation/TokenRequestValidator.cs
+++ b/src/IdentityServer4/Validation/TokenRequestValidator.cs
@@ -597,7 +597,7 @@ namespace IdentityServer4.Validation
             {
                 _logger.LogTrace("Client provided no scopes - checking allowed scopes list");
 
-                var clientAllowedScopes = _validatedRequest.Client.AllowedScopes;
+                var clientAllowedScopes = new List<string>(_validatedRequest.Client.AllowedScopes);
                 if (!clientAllowedScopes.IsNullOrEmpty())
                 {
                     if (_validatedRequest.Client.AllowOfflineAccess)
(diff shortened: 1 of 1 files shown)


chrisowhite:
Yeah, yeah, I noticed. Already on it. ;-)

brockallen:
Yes, excellent catch. Thanks!

PR #1076, opened 2017-04-19
Fix for issue#1068. Redact client secret in events.

akshetty9:
Fix for issue#1068. Redact client secret in events.


leastprivilege:
thanks!

PR #1119, opened 2017-05-01
Fixed AddDeveloperSigningCredential on full framework

jamesfera:
Calling AddDeveloperSigningCredential does work not on the Full Framework target.  The private key components of RSAParameters struct are non-serializable and therefore JSON.net can not properly serialize them.  A custom contract resolver was added to fix this.


leastprivilege:
Good catch! thanks!

brockallen:
Look good, thanks!

Issue #1084, opened 2017-04-19
Consistent expiration handling


leastprivilege:
see #1025.

chrisowhite:
@leastprivilege What do you want to consider expired? Current time > expire time, or current time >= expire time (expire time being [issue time + lifetime])? If I know which direction to go I can help with this.

PR #1135, opened 2017-05-05
Consistent expiration handling

chrisowhite:
Current time > expire time = expired grant.
- Flip expression in DateTimeExtensions.HasExpired() to read like HasExceeded().
- Check expiration and require consent for expired consent. Also remove consent from store.
- Remove expiration check from DefaultGrantStore that was returning null and causing callers to return invalid error statuses rather than expired. Also allows expired consents, refresh tokens, and reference tokens to be removed from the store.
- Change reference token expiration check to use DateTimeExtensions.HasExceeded().
- Remove DefaultGrantStore expired grant "should not load" tests.
- Add test for expired consent.
- Update Expired_Reference_Token test.

#1084


leastprivilege:
Thanks. We will look into it for v2.

brockallen:
Thanks!

PR #1172, opened 2017-05-15
修复vs for mac打开问题，主要更改.csproj文件的TargetFrameworks标签

haoyk:


leastprivilege:
Well - this doesn't really solve the problem. IdentityServer builds for multiple targets.

The problem is that VS for Mac is buggy and doesn't understand that.

PR #1192, opened 2017-05-23
Enhanced docs

weitzhandler:
Changed docs to clarify that the `claimTypes` parameter refers to `UserClaims`.


leastprivilege:
thanks

Issue #1216, opened 2017-05-31
Duplicate scopes in JWT claims


asfyra:
I have created an ApiResource with a several Scopes:

`                new ApiResource
                {
                    Name = "TestName",
                    DisplayName = "Allow application to access your account",
                    ApiSecrets = new List<Secret>{new Secret("scopeSecret".Sha256())},
                    Scopes = new List<Scope>
                    {
                        new Scope(Constants.Scope, "Allow application to access your account"){Required = true },
                        new Scope("ScopeA", "Allow application to make A"),
                        new Scope("ScopeB", "Allow application to make B")
                    }
                }`

Now when a JWT is generated for a client that is allowed to request those two scopes, they are included twice in token claims

`{
  "nbf": 1496218768,
  "exp": 1496222368,
  "iss": "MyIdentityServer",
  "aud": [
    "MyIdentityServer/resources",
    "customAPI",
    "TestName"
  ],
  "client_id": "TestWebApp",
  "sub": "375433",
  "auth_time": 1496218762,
  "idp": "local",
  "acr": "3ohqhanw43vr3f25f5xek0s1",
  "culture_name": "el-GR",
  "jti": "398f0a6b7e0f60551f39cba8324a36eb",
  "scope": [
    "customAPI.read",
    "customAPI.write",
    "customAPI.read",
    "customAPI.write",
    "ScopeA",
    "ScopeB",
    "ScopeA",
    "ScopeB"
  ],
  "amr": [
    "pwd"
  ]
}`

asfyra:
Issue solved with [this](https://github.com/IdentityServer/IdentityServer4/pull/1217/commits/16d11bdfc5cc2d81d454f637a2deee6b30298870)

mosdevid:
Thanks! You saved me some time there.

Issue #1099, opened 2017-04-24
Duplicate scopes in QS UI?


brockallen:
https://stackoverflow.com/questions/43419140/why-am-i-seeing-duplicate-scopes-on-identityserver4s-consent-screen/43520548?noredirect=1#comment74105237_43520548

asfyra:
If you are using InMemoryResourcesStore then this [fix](https://github.com/IdentityServer/IdentityServer4/pull/1217/commits/16d11bdfc5cc2d81d454f637a2deee6b30298870) might solve your issue

brockallen:
Fixed. Thanks.

PR #1217, opened 2017-05-31
Removed duplicate scopes

asfyra:
When using InMemoryResourcesStore you may get duplicate scopes in several  ui screens and inside jwt scopes section also.

This is a fix for an [issue i opened](https://github.com/IdentityServer/IdentityServer4/issues/1216)

This fix may also apply to [this issue](https://github.com/IdentityServer/IdentityServer4/issues/1099) if InMemory store is used. It solved the issue in my case.


brockallen:
Before I merge this, I'd like to repro the scenario (when I get time). Thanks for the PR.

brockallen:
Would you mind re-testing your scenario on the latest dev branch version? I don't currently see this behavior.

brockallen:
Ok, thanks for bringing this to our attention. I fixed it on dev.

Issue #1246, opened 2017-06-12
Consider better UTC now helper


brockallen:
As of now, it's static and sometimes our tests step on one another.

brockallen:
@leastprivilege thoughts? It's actually quite pervasive to make this change -- think all our date/time extension methods, and principal create static helper methods, etc.

brockallen:
merged

PR #1279, opened 2017-06-24
Design a better UTC now helper

brockallen:
Addresses #1246 // @leastprivilege have a look and see if you think it's worth it.


leastprivilege:
:shipit:

PR #1317, opened 2017-07-11
slightly cleaner GetOrigin implementation

smcl:
I noticed that `StringExtensions.GetOrigin` is a little hard to follow, and uses `String.IndexOf` and substring to approximate URL parsing. While it works alright, I think it's probably better to use the `Uri` class.

### DIFF
--- a/src/IdentityServer4/Extensions/StringsExtensions.cs
+++ b/src/IdentityServer4/Extensions/StringsExtensions.cs
@@ -211,17 +211,12 @@ namespace IdentityServer4.Extensions
 
         public static string GetOrigin(this string url)
         {
-            if (url != null && (url.StartsWith("http://") || url.StartsWith("https://")))
+            if (url != null)
             {
-                var idx = url.IndexOf("//", StringComparison.Ordinal);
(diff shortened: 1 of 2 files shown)


leastprivilege:
Thanks!

Coud you please add some tests to it?

smcl:
Good call - I have a set of addresses I tested this against already (taking the output of existing implementation vs mine) but I'm not sure what the best way to add these are. Perhaps I should add a separate pull req with the tests (so we have a set of inputs we know pass against the existing implementation) - then it's possible to test for regressions in the new implementation? What do you think? I know it's not a hugely important part of IdentityServer, but I'd like to get it right :-)

leastprivilege:
Either a new PR or add a commit. thanks

brockallen:
Last chance to add tests. Thanks.

smcl:
Hey Brock - sorry I have a lot of very poor excuses for not responding! I've added the tests that I put together - testing various combinations of scheme * host * port * resource against the value returned by the original implementation.

brockallen:
Cool, thanks!

PR #1326, opened 2017-07-13
added "alg" to JsonWebKey and DiscoveryResponseGenerator

filipw:
Even though it's optional, I though perhaps it can be added out of the box (see https://tools.ietf.org/html/rfc7517#section-4.4)
If this doesn't belong to the core server, my apologies - please just ignore the PR.

### DIFF
--- a/src/IdentityServer4/Models/JsonWebKey.cs
+++ b/src/IdentityServer4/Models/JsonWebKey.cs
@@ -14,5 +14,6 @@ namespace IdentityServer4.Models
         public string e { get; set; }
         public string n { get; set; }
         public string[] x5c { get; set; }
+        public string alg { get; set; }
     }
 }
--- a/src/IdentityServer4/ResponseHandling/DiscoveryResponseGenerator.cs
+++ b/src/IdentityServer4/ResponseHandling/DiscoveryResponseGenerator.cs
@@ -280,6 +280,7 @@ namespace IdentityServer4.ResponseHandling
         {
             var webKeys = new List<Models.JsonWebKey>();
             var signingCredentials = await Keys.GetSigningCredentialsAsync();
+            var algorithm = signingCredentials?.Algorithm ?? Constants.SigningAlgorithms.RSA_SHA_256;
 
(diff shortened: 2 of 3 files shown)


leastprivilege:
Thanks!

leastprivilege:
Hey, 

was there a practical reason you added this? Or just because you can? ;)

leastprivilege:
see https://github.com/IdentityServer/IdentityServer4/issues/3532

filipw:
it's in PingIdentity by default so this was added for an easier migration/compat path (otherwise you need to use a custom `DiscoveryResponseGenerator` with a custom subclass of `IdentityServer4.Models.JsonWebKey`).

Maybe the best solution would be to make it not shown when it's not explicitly chosen and add an overload to the key registration to include the signing algorithm to allow explicit choice.

leastprivilege:
What would happen if we remove it again?

leastprivilege:
I mean to you?! :)

filipw:
nothing really, if you consider this a blocker please remove it - I'd just need to add a custom `DiscoveryResponseGenerator` with a custom subclass of `IdentityServer4.Models.JsonWebKey` 😀

although I am not convinced this is the best idea, wouldn't it make more sense to publish `alg` on the response when it's known and just remove it when it's unknown (instead of falling back to RSA_SHA_256)?

leastprivilege:
My real question is - why do you need it?

The client needs to resolve a key from a kid. The real alg used to sign the token is inside the token? Why the extra alg on the discovery doc?

filipw:
> The client needs to resolve a key from a kid.

While `alg` is optional in the spec, isn't pretty much everything in JWKS is optional? Even `kid` is optional, as you could locate the key in the JWKS doc based on `x5t` instead. The way I interpret it, since there is nothing that mandates that `kid` is globally unique (or cryptgraphically unique, it can be just an arbitrary name), I should locate the key by both pieces of info from the token, `kid` + `alg` as that would be more accurate?

To me it makes sense to include it because then it's compatible with others - Auth0 does it this way too, even [Google does it](https://www.googleapis.com/oauth2/v3/certs).

PR #1339, opened 2017-07-17
Adding new blog links

damienbod:
@leastprivilege 
Greetings Damien


damienbod:
Do I need to do something here?

The missing cake.exe has nothing to do with the single file I changed.

Log:
Could not find Cake.exe at C:\projects\identityserver4\tools\Cake\Cake.exe

Greetings Damien

leastprivilege:
thanks - i think the build error is unrelated.

PR #1354, opened 2017-07-22
automatically add store implementations to DI when adding cached stores

duyvan82:
Currently, in order to add `ClientStoreCache` and `ResourceStoreCache`, we would have to do

```cs
services.AddTransient<CustomStoreCache>();
services.AddTransient<CustomResourceStoreCache>();

services.AddIdentityServer()
    .AddInMemoryCaching()
    .AddClientStoreCache<CustomStoreCache>()
    .AddResourceStoreCache<CustomResourceStoreCache>();
```

this pull request would help simplify the process by attempting to add the custom implementation of the stores within the `AddClientStoreCache<T>` and `AddResourceStoreCache<T>` extension methods, so that the above can be as simple as.

```csharp
services.AddIdentityServer()
    .AddInMemoryCaching()
    .AddClientStoreCache<CustomStoreCache>()
    .AddResourceStoreCache<CustomResourceStoreCache>();
```


brockallen:
Yep, looks good. Thanks.

PR #1357, opened 2017-07-24
Quickstart overview update

StevenTCramer:
Updated screen shots to reflect latest versions of Identity Server.  Cleaned up small wording issues. Updated to latest template code.  Like leaving in the `if (env.IsDevelopment())`and now UseApplicationInsights is in the default.

I removed the line:
> "In fact, when you try it, you should see an exception at startup time stating that services are missing."  

because the exception doesn't actually say that.

built the docs using "read the docs". Can be reviewed at the following:
http://identityserver4test.readthedocs.io/en/latest/quickstarts/0_overview.html


docs/quickstarts/0_overview.rst
```diff
@@ -107,6 +111,7 @@ we use the following configuration for the IdentityServer host in the quickstart
                 .UseContentRoot(Directory.GetCurrentDirectory())
                 .UseIISIntegration()
                 .UseStartup<Startup>()
+                .UseApplicationInsights()
```
brockallen: please remove this

brockallen:
looks good, other than the app insights -- remove that one line, please. pizza pizza!

docs/quickstarts/0_overview.rst
StevenTCramer: Can do.

StevenTCramer: removed :)

StevenTCramer:
Updated.  And thanks for the Pizza Pizza :)

leastprivilege:
thanks!

PR #1362, opened 2017-07-25
Logging instructions

nomad3k:
A couple of corrections to the logging documentation.
1. Indent the code so that it appears as a block, where presently it gets displayed on single line
1. Added two missing nuget packages.


leastprivilege:
thanks!

Issue #1066, opened 2017-04-17
Deal with Azure AD federation gateway problem


leastprivilege:
https://github.com/mclark1129/IdentityServer4.Samples/tree/master/AzureAd_CustomStateDataFormat

brockallen:
Done, but needs review/cleanup.

phrazed:
Any idea how to use these extension methods in Core 2? app.UseOpenIdConnectAuthentication is depreciated...

PR #1387, opened 2017-08-04
add state data formatter for OIDC state param #1066

brockallen:


brockallen:
Will do differently for AspCore 2

PR #1392, opened 2017-08-06
Show Default in comment/docs for booleans

StevenTCramer:
RequirePkce defaults to false.  I updated the comment to reflect that.


StevenTCramer:
I went ahead and did all the booleans in this file. and wrapped them like <c>true</c> or <c>false</c>
comment only changes.

brockallen:
Thanks!

PR #1409, opened 2017-08-12
AuthenticateAsync do not return null anymore

pksorensen:
AuthenticateAsync do not return null anymore (did not remove the null check as I dont know if it is a guarantied that it dont return null)


brockallen:
Looks like I accidentally added this check today. Feel free to have a look.

pksorensen:
no worries. Glad its there :)

PR #1414, opened 2017-08-15
include username in ExtensionGrantValidationContext

duyvan82:
a small pull request to simplify extension grant validation, specifically with grants that validates custom credentials and require access to the `ExtensionGrantValidationContext.Request.Username` property

currently we would have to read the username from `ExtensionGrantValidationContext.Request.Raw.Get("username")`, this pull request would make it possible to simply do `ExtensionGrantValidationContext.Request.Username`

### DIFF
--- a/src/IdentityServer4/Validation/ExtensionGrantValidator.cs
+++ b/src/IdentityServer4/Validation/ExtensionGrantValidator.cs
@@ -8,6 +8,8 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using IdentityModel;
+using IdentityServer4.Extensions;
 
 namespace IdentityServer4.Validation
 {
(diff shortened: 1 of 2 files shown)


duyvan82:
not sure why the travis build failed, looks like it's trying to build against netcore2.0? I didn't make any changes to the csproj files..

leastprivilege:
But having a username in an extension grant is a pretty niche use case - wouldn't you agree?

duyvan82:
@leastprivilege generally, yes I would, I just came across this with one of my extension grants with which I needed access to the username and thought it would make it easier if it was readily available in the context. It is a nice to have thing given the `Username` property is accessible from the `ExtensionGrantValidationContext`.

leastprivilege:
>  I just came across this with **one** of my extension grants with which I needed access to the username

That's exactly my point - with extension grants the client can send **anything** - I don't really like to special case the user name.

duyvan82:
@leastprivilege I get your point, having said that, how about we attempt to pre-populate other fields as well and not just username?
great talk at NDC {Sydney} today, I really enjoyed it, looking forward to the Authorization stuff *thumbsup*

leastprivilege:
Thanks!

I'm gonna close this for now. We can re-visit this post 2.0 maybe.

PR #1424, opened 2017-08-17
Remove Microsoft.AspNet.WebApi.Client dependency

Soarc:
This dependency is used in only one place to parse query parameters and can be easily avoided. 

Also this lib works only for net45 and this generates annoying warning.


leastprivilege:
thanks! This warning indeed annoyed me ;)

Issue #1514, opened 2017-09-12
Make Endpoint class public to allow custom routing


tillig:
I'm attempting to replace the discovery endpoint and possibly the endpoint router in an effort to provide a solution [similar to his Azure AD solution](http://www.cloudidentity.com/blog/2014/08/26/the-common-endpoint-walks-like-a-tenant-talks-like-a-tenant-but-is-not-a-tenant/) where tenant is a URL parameter.

I'm running into a challenge, though, with the fact the `IdentityServer4.Hosting.Endpoint` class is internal - in order to replace the "Discovery" endpoint, I have to either do some interesting reflection (so I can look in the `IServiceCollection` and find the `Endpoint` singleton that's already been registered) or craft my own `IIdentityServerBuilder` _before_ calling `services.AddIdentityServer()`. This is because registration _order_ of the endpoints appears to matter - the first one found in the service collection that matches the path is used, and that is always the default one rather than mine.

I also want to try introducing some route template support into a custom `IEndpointRouter` but, again, I can't get to the list of registered `Endpoint` objects very easily because `Endpoint` is internal.

**Any way `Endpoint` can be made public before the final 2.0 release?**

tillig:
Somewhat related, though I can open a different issue if desired: The `EndpointRouter` specifically has [a test to ensure the first registered route mapping is used](https://github.com/IdentityServer/IdentityServer4/blob/dev/test/IdentityServer.UnitTests/Hosting/EndpointRouterTests.cs#L85), which effectively _also_ ensures no one can override any of the default routes since the conforming DI container behavior is to return the `IEnumerable<T>` in the order of registration. I think in .NET Core this should be _finds the last registered mapping_ so overrides are possible.

tillig:
Looks like @brockallen had [an unanswered comment on 1364](https://github.com/IdentityServer/IdentityServer4/issues/1364#issuecomment-320474162) wondering if routes should be replaceable. I think so, and it'd be easy enough to do that.

Technically a custom `IEndpointRouter` could bypass use of any existing registered `Endpoint` objects and re-register everything in a custom way to do it, but it'd be nice to be able to get the same effect with less work.

tillig:
[...]

PR #1515, opened 2017-09-12
Made Endpoint public so it can be used in custom routing.

tillig:
Addresses issue #1514.


brockallen:
Yep, this was just an oversight. Thanks

Issue #1518, opened 2017-09-13
Consider ISystemClock?


leastprivilege:

brockallen:
Done. Little gain, IMO, but done.

PR #1524, opened 2017-09-14
use ISystemClock

brockallen:
#1518


leastprivilege:
OK

filipw:
Sorry for commenting on an old PR, but what is the purpose/benefit of this change?

brockallen:
Testing following ASP.NET Core's approach

Issue #1534, opened 2017-09-18
Consider making client claims prefix value configurable


leastprivilege:
instead of boolean and default to "client_"

brockallen:
Done.

PR #1546, opened 2017-09-21
making client claims prefix value configurable #1534

AliBazzi:
Please take a look at the proposed implementation, please ignore if this isn't what you have in your mind.

Regards.


leastprivilege:
Thanks - we haven't discussed this yet - but I will have look.

leastprivilege:
thanks!

PR #1558, opened 2017-09-25
Add new Node API sample to community quickstarts

lyphtec:
Hi,

I created a new sample / quickstart on securing a Node API with IdentityServer4 & have added a link to it in the Community Quickstarts document.

This uses higher quality - production ready modules comparing to the one used in the NodeJsApi sample.


leastprivilege:
@AlexZeitler review?

AlexZeitler:
LGTM + suggestions:

* Put everything in Containers + Docker Compose as the current sample does to ease startup with a single command and without installing deps locally
* Use release version of IdSrv4 2.0 instead of RC
* Add script to create the signing cert as the current sample does (just use the script), maybe add a .cmd/.ps1 for Windows as well
* Integrate the JS Implicit sample client as the current sample does
* Current sample has auto discovery for `jwks` endpoint, maybe re-integrate this again

Good to see `jwt` / `jwks-rsa` Node modules from Auth0 now support x5t / x5c.

Tested on Ubuntu 16.04.

AlexZeitler:
BTW if suggestions are applied we should make this sample the "official" Node.js sample (read: replace the current `NodeJsApi` folder in `IdentityServer4.Samples`). 
Otherwise it would make sense to refactor that sample to use `jwt` / `jwks-rsa` as they now handle `x5t`/`x5c` as well.
The currently used Node.js module to handle this was more a proof of concept and it was just implemented because `jwt` / `jwks-rsa` did not support `x5t` / `x5c` at the time of implemeting the sample.
Instead of spreading efforts for Node.js samples we should combine them to have an evolution instead of many revolutions.

My rationale behind using Docker for the Node.js sample is: people in Node.js land are looking for an OIDC Server and you can just point them to the sample without having them to deal with .NET Core installation etc. to just get a look and play with it. Therefore also the suggestion to stick (or at least to re-add) with the OIDC Implicit Client.

lyphtec:
Hi,

Thanks for the suggestions.

I've updated the sample:

- Removed all reference to the current NodeJsApi sample in the README
- Added JS Implicit sample client
- Added Docker support. I've re-worked the samples to use env variables when running in Docker to change the IssuerUri - so this avoids having to mess around with the HOST_IP as per the NodeJsApi sample
- Updated to release version of IdSrv4 2.0.0
- Updated README with usage of self-sign cert & how user can potentially generate their own if they want - (otherwise, just use the included `cert.pfx` that I have re-generated with a lifetime of 10 years). Also included Docker usage section.

Best.

leastprivilege:
I don't have the time to look into that - @AlexZeitler is that the sample YOU would merge in?

AlexZeitler:
Will have a look the next few days but in general it sounds good to me now.

brockallen:
Update?

AlexZeitler:
Tomorrow, sorry for the delay.

leastprivilege:
I actually think we should remove all samples that are not written by ourselves - we can't maintain them. I'd rather link to the sample in our docs..

@brockallen  thoughts?

AlexZeitler:
Sample works as expected.

brockallen:
I added this manually since merge conflict up here aren't working. Thanks!

Issue #1572, opened 2017-09-27
Route matching not separate from endpoint route handling


tillig:
I've been working on a multitenant version of Identity Server where routing matches a template like `/{tenantid}/connect/token` or whatever - the tenant ID is a route parameter, similar to the way Azure AD does things. This wasn't hard to do with my own `IEndpointRouter`. Really nice flexibility around that.

A challenge I've hit is that even with the routing, the _endpoint handlers themselves_ are checking paths to see how to process certain requests:

- `AuthorizeEndpoint` looks at the request path [and tests for equality](https://github.com/IdentityServer/IdentityServer4/blob/74827bc4df39c2234d9490ba5ce032ce806ab475/src/IdentityServer4/Endpoints/AuthorizeEndpoint.cs#L56) to see if it's `/connect/authorize` or `/connect/authorize/callback`.
- `EndSessionEndpoint` also [tests the request path for equality](https://github.com/IdentityServer/IdentityServer4/blob/74827bc4df39c2234d9490ba5ce032ce806ab475/src/IdentityServer4/Endpoints/EndSessionEndpoint.cs#L36) to see if it's `/connect/endsession` or `/connect/endsession/callback`.

The path checking in the endpoint handler breaks the separation between routing and endpoint handling. Given the endpoint handlers are internal, you can't derive-and-override, either; it's a full fork/copy/paste situation.

I can think of a couple of solutions (not counting "fork/copy/paste"):

1. **Map one endpoint handler to each endpoint.** For handlers like `AuthorizeEndpoint` or `EndSessionEndpoint`, have a base class for common logic and individual derived classes for each matching route. This would avoid the need for the equality check.
2. **Make the endpoint handlers public so folks can derive/override the logic.** I can see why that may not be the best option, I'm just trying to avoid copy/pasting a bunch.

If you're open to option 1 I could PR that.

tillig:
Note the fork/copy/paste route is a bit painful due to so many things being internal. For example, `AuthorizeEndpoint`...

- Returns an `AuthorizeResult`, which is internal.
- Logs with `AuthorizeRequestValidationLog` and `AuthorizeResponseLog`, both of which are internal (and uses `LogSerializer`, which is also internal).
- Uses a lot of internal extension methods along the way (conversions from objects to `NameValueCollection`, etc.).

brockallen:
We'll discuss.

tillig:
[...]

PR #1576, opened 2017-09-28
Separate callback route endpoints from base route endpoints

tillig:
Addresses #1572:

- Split `/connect/authorize` endpoint from `/connect/authorize/callback`. Retains common code in abstract base class.
- Split `/connect/endsession` endpoint from `/connect/endsession/callback`. No common code was shared.

All tests were retained and still pass, though the `authorize` endpoint tests got split so they only run against the appropriate endpoint class.

### DIFF
--- a/src/IdentityServer4/Configuration/DependencyInjection/BuilderExtensions/Core.cs
+++ b/src/IdentityServer4/Configuration/DependencyInjection/BuilderExtensions/Core.cs
@@ -76,6 +76,7 @@ namespace Microsoft.Extensions.DependencyInjection
         {
             builder.Services.AddTransient<IEndpointRouter, EndpointRouter>();
 
+            builder.AddEndpoint<AuthorizeCallbackEndpoint>(EndpointNames.Authorize, ProtocolRoutePaths.AuthorizeCallback.EnsureLeadingSlash());
             builder.AddEndpoint<AuthorizeEndpoint>(EndpointNames.Authorize, ProtocolRoutePaths.Authorize.EnsureLeadingSlash());
             builder.AddEndpoint<CheckSessionEndpoint>(EndpointNames.CheckSession, ProtocolRoutePaths.CheckSession.EnsureLeadingSlash());
             builder.AddEndpoint<DiscoveryEndpoint>(EndpointNames.Discovery, ProtocolRoutePaths.DiscoveryConfiguration.EnsureLeadingSlash());
--- a/src/IdentityServer4/Configuration/DependencyInjection/BuilderExtensions/Core.cs
+++ b/src/IdentityServer4/Configuration/DependencyInjection/BuilderExtensions/Core.cs
@@ -79,6 +79,7 @@ namespace Microsoft.Extensions.DependencyInjection
             builder.AddEndpoint<AuthorizeCallbackEndpoint>(EndpointNames.Authorize, ProtocolRoutePaths.AuthorizeCallback.EnsureLeadingSlash());
             builder.AddEndpoint<AuthorizeEndpoint>(EndpointNames.Authorize, ProtocolRoutePaths.Authorize.EnsureLeadingSlash());
             builder.AddEndpoint<CheckSessionEndpoint>(EndpointNames.CheckSession, ProtocolRoutePaths.CheckSession.EnsureLeadingSlash());
+            builder.AddEndpoint<DiscoveryKeyEndpoint>(EndpointNames.Discovery, ProtocolRoutePaths.DiscoveryWebKeys.EnsureLeadingSlash());
             builder.AddEndpoint<DiscoveryEndpoint>(EndpointNames.Discovery, ProtocolRoutePaths.DiscoveryConfiguration.EnsureLeadingSlash());
             builder.AddEndpoint<EndSessionCallbackEndpoint>(EndpointNames.EndSession, ProtocolRoutePaths.EndSessionCallback.EnsureLeadingSlash());
             builder.AddEndpoint<EndSessionEndpoint>(EndpointNames.EndSession, ProtocolRoutePaths.EndSession.EnsureLeadingSlash());
--- a/src/IdentityServer4/Configuration/DependencyInjection/BuilderExtensions/Core.cs
+++ b/src/IdentityServer4/Configuration/DependencyInjection/BuilderExtensions/Core.cs
@@ -80,6 +80,7 @@ namespace Microsoft.Extensions.DependencyInjection
             builder.AddEndpoint<AuthorizeEndpoint>(EndpointNames.Authorize, ProtocolRoutePaths.Authorize.EnsureLeadingSlash());
             builder.AddEndpoint<CheckSessionEndpoint>(EndpointNames.CheckSession, ProtocolRoutePaths.CheckSession.EnsureLeadingSlash());
             builder.AddEndpoint<DiscoveryEndpoint>(EndpointNames.Discovery, ProtocolRoutePaths.DiscoveryConfiguration.EnsureLeadingSlash());
+            builder.AddEndpoint<EndSessionCallbackEndpoint>(EndpointNames.EndSession, ProtocolRoutePaths.EndSessionCallback.EnsureLeadingSlash());
             builder.AddEndpoint<EndSessionEndpoint>(EndpointNames.EndSession, ProtocolRoutePaths.EndSession.EnsureLeadingSlash());
             builder.AddEndpoint<IntrospectionEndpoint>(EndpointNames.Introspection, ProtocolRoutePaths.Introspection.EnsureLeadingSlash());
             builder.AddEndpoint<TokenRevocationEndpoint>(EndpointNames.Revocation, ProtocolRoutePaths.Revocation.EnsureLeadingSlash());
--- a/src/IdentityServer4/Endpoints/AuthorizeCallbackEndpoint.cs
+++ b/src/IdentityServer4/Endpoints/AuthorizeCallbackEndpoint.cs
@@ -1,0 +1,75 @@
+﻿// Copyright (c) Brock Allen & Dominick Baier. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
+
+using System;
+using System.Net;
+using System.Threading.Tasks;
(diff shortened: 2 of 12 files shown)


tillig:
This is more to facilitate the research and discussion over #1572. It wasn't a lot of work and I thought it may help to visualize what the actual changes required end up being.

All endpoint classes can still remain internal so no additional API footprint is exposed.

brockallen:
There are just those two that were handling multiple paths, right?

tillig:
That's correct, it was just those. The rest are fine as is.

brockallen:
Ok yes, I don't see any real issues with doing this. I'll get it in today.

tillig:
Woot! Thanks!

brockallen:
Oops, the discovery endpoint also handles two different paths.

tillig:
Oh, I missed that. No worries, I'll get that in for you Monday.

brockallen:
It looks like you're relying upon the order of registration in DI of endpoints to deal with overlapping endpoint paths (given the use of `StartsWithSegments`). Do you think we should change our route matching to be exact? Originally the use of `StartsWithSegments` was to allow an EP implementation to handle nested paths.

brockallen:
The more I think about it, I think I like to make it explicit -- one EP == one path, rather than using `StartsWithSegments`. It makes it very cut and dry (IOW, simple and deliberate).

tillig:
OK, I can update the router to use equality.

For reference, in my custom router I used MVC routing template matching, which is equality... But I also stored route data in the request context for additional extensibility. That would be an easy enhancement later if path matching was equality.

Anyway, I'll update for equality.

brockallen:
There's a test that check for nesting - we'll want to change it to prohibit nesting :)

brockallen:
This also makes me wonder -- we now have the possibility of overrides, meaning someone else can put a route in DI and suppress the default ones. 

For your multi-tenancy scenario, are you overriding our built-in routes? Or just adding new ones to the pipeline?

tillig:
I end up overriding them. I resolve the set of endpoints and create a new set based on the current settings and then adding the route parameter prefix.

Give me about an hour and I'll show you the code. (I don't have it on my phone.)

tillig:
OK, for completeness sake I'll start from the top.

The use case is multitenancy in a pattern like Azure AD: `https://identityserver/{tenantid}/connect/token` where the tenant ID is the route parameter there.

We created our own endpoint type that uses stuff from the `Microsoft.AspNetCore.Routing.Template` namespace.

```csharp
public class TemplateEndpoint
{
  private string _template;

  public Type Handler { get; set; }

  public string Name { get; set; }

  public int Order { get; set; }

  public TemplateMatcher Matcher { get; private set; }

  public string Template
  {
    get
    {
      return this._template;
    }

    set
    {
      if (value == null)
      {
        this._template = null;
        this.Matcher = null;
      }
      else
      {
        this._template = value;
        this.Matcher = new TemplateMatcher(TemplateParser.Parse(value), new RouteValueDictionary());
      }
    }
  }

  public virtual bool Match(HttpContext context)
  {
    if (context == null)
    {
      throw new ArgumentNullException(nameof(context));
    }

    if (this.Matcher.TryMatch(context.Request.Path, values))
    {
      // Sets the dictionary of route values in HttpContext so it can be used later.
      context.SetRouteValues(values);
      return true;
    }

    return false;
  }
}
```

The important bits there are:

- Endpoints have an `Order` so the order of registration is not important. That was a defect in the assumptions used in Microsoft.Extensions.DependencyInjection that we raised a lot of stink over in the early days but there were so many things in ASP.NET Core that assumed resolution would maintain order of registration that we couldn't fix it.
- Endpoints have a `Match` method so technically we can break that up later and have different endpoint types that match based on different criteria and that's abstracted away from the router. The router is only interested in connecting endpoints with inbound requests.

We also have a `PathEndpoint` type that basically is the same as the internal `IdentityServer4.Hosting.Endpoint` type. We use reflection during our custom routing setup to re-register the internal routes as types we can use. We won't need that once the `Endpoint` type is public.

```csharp
public static IIdentityServerBuilder AddMultitenantRouting(this IIdentityServerBuilder builder)
{
  if (builder == null)
  {
    throw new ArgumentNullException(nameof(builder));
  }

  builder.Services.AddTransient<IEndpointRouter, MultitenantEndpointRouter>();

  // Map IdentityServer internal endpoints to custom endpoint structures
  // that match based on path segments. The handler will have already
  // been registered with the service collection so we only need to map
  // the endpoint structures.
  foreach (var descriptor in builder.Services.Where(sd => sd.ServiceType == IdentityServerEndpointType).ToArray())
  {
    dynamic originalEndpoint = new TransparentObject(descriptor.ImplementationInstance);
    var newEndpoint = new PathEndpoint
    {
      Handler = (Type)originalEndpoint.Handler,
      Name = (string)originalEndpoint.Name,
      Order = 100,
      Path = (PathString)originalEndpoint.Path,
    };
    builder.Services.AddSingleton(newEndpoint);
  }

  return builder;
}
```

The final piece of the puzzle is the `MultitenantEndpointRouter`. This is where we take the currently registered path-based endpoints and update them to understand route templates.

```csharp
public class MultitenantEndpointRouter : IEndpointRouter
{
  private readonly IConfiguration _configuration;

  private readonly IEnumerable<TemplateEndpoint> _endpoints;

  private readonly ILogger<MultitenantEndpointRouter> _logger;

  private readonly IdentityServerOptions _options;

  public MultitenantEndpointRouter(
    IEnumerable<PathEndpoint> endpoints,
    IdentityServerOptions options,
    IConfiguration configuration,
    ILogger<MultitenantEndpointRouter> logger)
  {
    if (endpoints == null)
    {
      throw new ArgumentNullException(nameof(endpoints));
    }

    if (options == null)
    {
      throw new ArgumentNullException(nameof(options));
    }

    if (configuration == null)
    {
      throw new ArgumentNullException(nameof(configuration));
    }

    if (logger == null)
    {
      throw new ArgumentNullException(nameof(logger));
    }

    // Templates will be like
    // {tenantid}/connect/token/{*catchall}
    // Also, here's where we order things by that `Order` property.
    this._endpoints = endpoints
      .OrderBy(e => e.Order)
      .Select(e => new TemplateEndpoint { Handler = e.Handler, Name = e.Name, Order = e.Order, Template = $"{{tenantid}}{e.Path}/{{*catchall}}" })
      .ToArray();
    this._options = options;
    this._configuration = configuration;
    this._logger = logger;
  }

  public IEndpointHandler Find(HttpContext context)
  {
    if (context == null)
    {
      throw new ArgumentNullException(nameof(context));
    }

    foreach (var endpoint in this._endpoints)
    {
      if (endpoint.Match(context))
      {
        // Extension method to read a previously identified tenant
        // out of context route data.
        var tenant = context.GetTenantIdRouteValue();
        if (tenant == null ||
          (tenant.Equals("common", StringComparison.Ordinal) && !endpoint.Name.Equals("Discovery", StringComparison.OrdinalIgnoreCase)))
        {
          // If there is NO tenant specified OR
          // The tenant is "common" but the endpoint isn't just the disco endpoint
          // Then we can't handle the request.
          // In v1 we only allow common requests for metadata. We may later open
          // up tenant-agnostic requests for tokens like Azure AD.
          //
          // The ID "Discovery" is defined in the internal IdentityServer4.Constants.EndpointNames.Discovery
          // constant but is internal.
          return null;
        }
        else if (!tenant.Equals("tenantid", StringComparison.Ordinal))
        {
          try
          {
            // Extension method to read the configuration
            // for a tenant given a tenant ID. It throws if the
            // tenant isn't found.
            this._configuration.GetTenant(tenant);
          }
          catch (InvalidOperationException)
          {
            // Tenant not found.
            return null;
          }
        }

        // Set the tenant ID on the context in a standard/central spot for later
        // use by tenant identification strategies and so on. Convenience around
        // context.Items access.
        context.SetTenantId(tenant);

        this._logger.LogDebug("Request path {path} matched to endpoint type {endpoint}", context.Request.Path, endpoint.Name);

        // Update the base path so it includes the tenant. That fixes metadata doc
        // generation and so on.
        context.SetIdentityServerBasePath(context.GetIdentityServerBasePath() + "/" + tenant);
        return this.GetEndpointHandler(endpoint, context);
      }
    }

    this._logger.LogTrace("No endpoint entry found for request path: {path}", context.Request.Path);

    return null;
  }

  private IEndpointHandler GetEndpointHandler(Endpoint endpoint, HttpContext context)
  {
    if (this._options.Endpoints.IsEndpointEnabled(endpoint))
    {
      var handler = context.RequestServices.GetService(endpoint.Handler) as IEndpointHandler;
      if (handler != null)
      {
        this._logger.LogDebug("Endpoint enabled: {endpoint}, successfully created handler: {endpointHandler}", endpoint.Name, endpoint.Handler.FullName);
        return handler;
      }
      else
      {
        this._logger.LogDebug("Endpoint enabled: {endpoint}, failed to create handler: {endpointHandler}", endpoint.Name, endpoint.Handler.FullName);
      }
    }
    else
    {
      this._logger.LogWarning("Endpoint disabled: {endpoint}", endpoint.Name);
    }

    return null;
  }
}
```

The usage ends up looking like:

```csharp
services
  .AddIdentityServer(opt => { })
  .AddMultitenantRouting();
```

That's a lot to get through, but there you go.

For now, the real blockers in the published version of IdentityServer are:

- `Endpoint` is internal.
- Endpoint handlers are doing path matching internally and taking control away from the routing facility.

The `Endpoint` has been updated to be public, so that's done. Now if we can get the handlers to stop path matching, I'll be flying.

In the meantime, I can update the routing matching to be path equality and I'll find that test that checks for nesting and fix that up, too.

tillig:
Note I think I can remove that `{*catchall}` bit from my template-based route matching. I added that when endpoints were handling multiple paths but can probably remove it now for a more exact matching.

tillig:
OK, updated for routing to use exact path match, including that unit test.

tillig:
If you want, I can switch it to use the templating thing like I have. You wouldn't need any route parameters by default, so it'd be exact path match, but custom routing could update the templates as needed.

brockallen:
No, I think this is sufficient if you're satisfied.

tillig:
Works for me - thanks!

PR #1578, opened 2017-10-01
Implemented PairWise Subject

DerAlbertCom:
I've implemted Pairwise Subject. Not ready for merge yet. I made this PR to get early feedback.

1. Creation of the Pairwise Subject in DefaultClaimService (SubjectId+Salt).Sha256(), to not break backward compatiblity with the virtual method GetStandardSubjectClaimsAsync() i had to transform the Subject Id claim bevor adding them.
2. EndSessionRequestValidator changed to validate the new PairwiseSubject if needed
3. Wrote Unit Tests for all that.
4. Changed DiscoveryResponseGenerator to announce the supported subject types.
5. Adapted Integrationtest for that

ToDo:
1. Write Integrationtests for Pairwise Subject
2. May create IClientSubjectService for creation and valdiation of the client subject, to seperate the concern of creating and validating the subject which is delivered to the client.
```
public interface IClientSubjectService {
     string GeneratedSubject(string userSubject, Client client);
     bool ValidateSubject(string userSubject, string clientSubject, Client client); 
}
```

### DIFF
--- a/IdentityServer4.v3.ncrunchsolution
+++ b/IdentityServer4.v3.ncrunchsolution
@@ -1,0 +1,6 @@
+﻿<SolutionConfiguration>
+  <Settings>
+    <AllowParallelTestExecution>True</AllowParallelTestExecution>
+    <SolutionConfigured>True</SolutionConfigured>
+  </Settings>
+</SolutionConfiguration>
--- a/src/IdentityServer4/Configuration/DependencyInjection/BuilderExtensions/Core.cs
+++ b/src/IdentityServer4/Configuration/DependencyInjection/BuilderExtensions/Core.cs
@@ -147,6 +147,7 @@ namespace Microsoft.Extensions.DependencyInjection
             builder.Services.TryAddTransient<ITokenService, DefaultTokenService>();
             builder.Services.TryAddTransient<ITokenCreationService, DefaultTokenCreationService>();
             builder.Services.TryAddTransient<IClaimsService, DefaultClaimsService>();
+            builder.Services.TryAddTransient<IClientSubjectService, DefaultClientSubjectService>();
             builder.Services.TryAddTransient<IRefreshTokenService, DefaultRefreshTokenService>();
             builder.Services.TryAddTransient<IConsentService, DefaultConsentService>();
             builder.Services.TryAddTransient<ICorsPolicyService, DefaultCorsPolicyService>();
--- a/src/IdentityServer4/Constants.cs
+++ b/src/IdentityServer4/Constants.cs
@@ -88,7 +88,7 @@ namespace IdentityServer4
             OidcConstants.ResponseModes.Fragment
         };
 
-        public static string[] SupportedSubjectTypes =
+        public static readonly string[] SupportedSubjectTypes =
         {
             "pairwise", "public"
         };
--- a/src/IdentityServer4/ResponseHandling/DiscoveryResponseGenerator.cs
+++ b/src/IdentityServer4/ResponseHandling/DiscoveryResponseGenerator.cs
@@ -241,7 +241,7 @@ namespace IdentityServer4.ResponseHandling
                 entries.Add(OidcConstants.Discovery.TokenEndpointAuthenticationMethodsSupported, SecretParsers.GetAvailableAuthenticationMethods().ToArray());
             }
 
-            entries.Add(OidcConstants.Discovery.SubjectTypesSupported, new[] { "public", "pairwise" });
+            entries.Add(OidcConstants.Discovery.SubjectTypesSupported, Constants.SupportedSubjectTypes);
             entries.Add(OidcConstants.Discovery.IdTokenSigningAlgorithmsSupported, new[] { Constants.SigningAlgorithms.RSA_SHA_256 });
             entries.Add(OidcConstants.Discovery.CodeChallengeMethodsSupported, new[] { OidcConstants.CodeChallengeMethods.Plain, OidcConstants.CodeChallengeMethods.Sha256 });
 
--- a/src/IdentityServer4/ResponseHandling/DiscoveryResponseGenerator.cs
+++ b/src/IdentityServer4/ResponseHandling/DiscoveryResponseGenerator.cs
@@ -241,7 +241,7 @@ namespace IdentityServer4.ResponseHandling
                 entries.Add(OidcConstants.Discovery.TokenEndpointAuthenticationMethodsSupported, SecretParsers.GetAvailableAuthenticationMethods().ToArray());
             }
 
-            entries.Add(OidcConstants.Discovery.SubjectTypesSupported, new[] { "public" });
+            entries.Add(OidcConstants.Discovery.SubjectTypesSupported, new[] { "public", "pairwise" });
             entries.Add(OidcConstants.Discovery.IdTokenSigningAlgorithmsSupported, new[] { Constants.SigningAlgorithms.RSA_SHA_256 });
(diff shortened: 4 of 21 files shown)


DerAlbertCom:
Working on Integrations Tests. Now it is tricky ;) The DefaultCustomTokenValidator fails, because of the pairwise subject. Have to think about it, any hints welcome.

DerAlbertCom:
Ok, on the first sight it seems not possible to implement this without breaking changes. Breaking on public interfaces  (which are currently internal implemented like ITokenValidator, IIntrospectionRequestValidator, maybe more). And alternative may be using IHttpContextAccessor to get the current principal on several places.

DerAlbertCom:
After more changed integration test (adding PairWiseSalt to some current Tests), i have it up an running again. They main problem was when the AccessToken is a reference token. The generation of the AccessToken included the PairWise Subject (which made no sense, because the token is only stored locally, because the reference is transfered to the client).

I introduced the IClientSubjectService with DefaultClientSubjectService. And hat two introduced to small _**breaking changes.**_ In the constructors of the *DefaultClaimsService* and *DefaultCustomTokenValidator*, they get additional parameters.

* DefaultClaimsService  gets additionally the IClientSubjectService
  _it is possible to not use the IClientSubjectService and have no breaking change. But then the subject generation is not customizable._
* DefaultCustomTokenValidator gets additionally the IHttpContextAccessor for accessing the current logged in user, to avoid breaking changes in the ITokenValidator interface. 
  _It may possible to make an additional method overloads at the interface which also has a ClaimsPrincipal parameters for additional validation. But technically these will lead to not uses old method interface definitions. _

I think there must be more integration tests be generated.

leastprivilege:
Hey, 

thanks for the work so far.

We are still planning to include that in 2.1 - but we need to discuss how we deal with the breaking changes.

If you want, add more tests, that will make it easier for us.

TomCJones:
I am ready to build and test pairwise now. What is status vis.a.vis the dev branch?

brockallen:
@TomCJones it's a breaking change, so we won't include this until 3.x. also, we have another issue related to this that needs to be addressed/solved.

TomCJones:
what change is that?
I may go ahead an impalement if you have some hints about where you will be heading.

brockallen:
> what change is that?

we need to make the sid unique per client. otherwise pairwise sub is not as useful as it's intended to be.

TomCJones:
That is a very interesting observation, I have never heard a threat that includes correlations of SIDs. I suppose it is possible, but can't really picture the exploit. Do you have a specific exploit that you have heard. I would like to understand so it could be captured.

Session ID - String identifier for a Session. This represents a Session of a User Agent or device for a logged-in End-User at an RP. Different sid values are used to identify distinct sessions at an OP. The sid value need only be unique in the context of a particular issuer. Its contents are opaque to the RP. Its syntax is the same as an OAuth 2.0 Client Identifier.

brockallen:
No, it's just something we considered as we were thinking about this PR. sid is not the same as sub, but it's a type of correlation.

TomCJones:
well, if you don't mind I will bring it up in the openid list to see if there are others that share your concern

brockallen:
Sure, feel free to. I'm on the list, so I'll be interested in the replies.

brockallen:
BTW, my plan was to gen the per-client sid from the "real" sid and do something like: client_sid=Sha256(client_id + sid). I think this will work fine in the check_session logic, and it's easy to re-gen that at both front and back channel signout.

TomCJones:
Nat and I talked about the SID at a FAPI meeting today. It seems that the best way to address issues with the sid is to make it short lived, which is what Amazon does. Of course their is no downside to your suggestion either.  I collected information on sid's of all sorts on this site as I was exploring the problem. https://wiki.idesg.org/wiki/index.php?title=Session_Identifier

brockallen:
What would short-lived look like or mean?

TomCJones:
https://stackoverflow.com/questions/25686484/what-is-intent-of-id-token-expiry-time-in-openid-connect

thx ..Tom (mobile)

On Apr 3, 2018 5:27 PM, Brock Allen <notifications@github.com> wrote:

What would short-lived look like or mean?

brockallen:
That SO link is unrelated to my question for you. I'm talking about the sid.

brockallen:
> So if the ID token expires would any new id token not have a new sid

Clients might have ditched the id_token after initial validation. They tend to keep the sid as separate state in the client app (e.g. as a claim in the user's session cookie). That's what's then used to validate the front-channel SLO notification. This means the sid issued at sign-in via the id_token must be the same used at SLO time.

TomCJones:
OK, so it's an interop question that is not disambiguated by the spec. I guess I now have a question I can ask as an ab/connect issue.  (As an aside, the few places where the sid appears would make a low priority privacy issue.)

brockallen:
Well, I think the spec is clear. When the initial id_token is sent it has a sid. That is the same sid that will be passed at front/back channel signout notification.

TomCJones:
If that is the case then the session id is what the OP understand it to be and that may have no relationship to what the RP thinks the session is. That makes this definition, from the spec, wrong:
Session
Continuous period of time during which an End-User accesses a Relying Party relying on the Authentication of the End-User performed by the OpenID Provider.

brockallen:
I guess I'm missing why it's so complicated.

A user has a session at the STS. The STS sends a sid to a client (via id_token). When the user's session at the STS ends, the STS notifies the client, passing the sid. 

The client manages its own session for the user, but maintains the sid as an associated value with the user's session at the client. When the STS notifies the client with the sid, the client checks if the sid associated with the current user's session matches, and now the client knows the user's session at the STS is ended. It decides what it wants to do next (e.g. sign the user out, prompt the user, ignore the notification, whatever).

I don't see any reason the STS must use the same sid across multiple clients, but it must use the same sid for an individual client.

Is my thinking wrong? If not, then does the wording of the spec disagree with what I described?

TomCJones:
That is not a flow that i have ever used. Nor is it one that i plan to use. If that is the only flow the logout documents support, then i guess i would have no use for them or for sid. Sorry for the distraction.

brockallen:
> That is not a flow that i have ever used

How did you (or do you) think sessions and signout worked? Or are you not building web apps?

TomCJones:
Sure, with oidc the sessions can (& imho should) be created asynchronously. So given the requirements of GDPR the initial session between the user and the RP should have the minimum claims known to be required. If at a later time the user decides to release additional claims to the RP from the same or a different OP, the original OP claims are retained and new ones added. Per security requirements (eg owasp) a new session with the RP must be established, which doesn't necessarily require a new one with the original OP.

brockallen:
Ah I think I understand why we're talking past one another -- I don't think of a session at the client by being identified by a unique id (necessarily). I normally just think of a cookie with claims, and that cookie is often renewed/slid. So to be going back to the OP for more/new claims... I don't know if that really is a new session at the client.

TomCJones:
consider fapi pt 2 vs fapi pt 1. If I signin as with part 1 security I may have a session that can be attacked. If I decide to transfer money and move the connection to fapi pt2 I definitively do not want the old session (which may have been subverted) to be used.  I believe that is why owasp REQUIRES a new session if new claims are included. Perhaps that is too strong, but it is probably better to tell devs NEVER to use the same session if any condition of the original session is altered. That is the safer policy.

Peace ..tom
________________________________

DerAlbertCom:
i close this pr because it is postponed to IdServer4 V3

PR #1598, opened 2017-10-06
Set NoCache headers for discovery document if CacheInterval is set to 0

NilsEngelbach:
To avoid caching of the discovery document the option ResponseCacheInterval is set to 0
```
options.Discovery.ResponseCacheInterval = 0
```
Expected behavior would be that the the NoCache headers are set correctly.
Current behavior no cache control header at all.


brockallen:
thanks

PR #1604, opened 2017-10-09
Moving IdentityServerMiddleware IEventService parameter from constructor to Invoke

cruikshj:
- The goal is to allow per-request instances of IEventSink.
 - Attempting to use per-request (scoped) instances of IEventService/IEventSink results in a runtime exception due to the fact that IdentityServerMiddleware injects an IEventService during construction, rather than per-request.


brockallen:
Yes, good catch. That's a bug, for sure.

leastprivilege:
thanks!

PR #1621, opened 2017-10-13

wstllc:
I did not change the image that would now look slightly different if the user were to implement the proposed code for Program.cs... if this PR is accepted, perhaps that image ought to be updated as well.


wstllc:
I just realized a small change would make my code submission better...  adjust comments in that code I submitted as shown below...

// password grant
// **NOTE: the next two lines of code are explained in upcoming section titled Protecting an API using Passwords**
            var client = new TokenClient(disco.TokenEndpoint, "ro.client", "secret");
            return await client.RequestResourceOwnerPasswordAsync("alice", "password", "api1");

leastprivilege:
Thanks! But I simply added a link to the client source code in the samples repo.

PR #1646, opened 2017-10-19
Fix for One-Time refresh token not being removed from PersistedGrants store

eldarkra:
The TokenResponseGenerator passes an uninitialized RefreshTokenHandle to the DefaultTokenService when requesting a refresh token update.
To solve this issue, the RefreshTokenHandle is being stored on the ValidatedRequest object.


leastprivilege:
Thanks! - I already did that due to another problem.

PR #1702, opened 2017-11-03
Patch 2

Triwaters:
Apologies if I am doing this improperly. As noted in prior communications, I'm rather new to most of this. I'm confident, however, that the proposed fix  is a necessary one.


brockallen:
Yep, looks good. Thanks!

PR #1705, opened 2017-11-03
per email to Brock, fix incorrect documentation

Triwaters:
Again, could be mistaken. My experience was that IdentityServer4 did *not* come along for the ride.


Triwaters:
my mistake (no shock). No change required to documentation here.

PR #1706, opened 2017-11-03
Fix #1697

### LINKED ISSUE #1697: Parameters twisted for GetConsentKey
Hey guys,

on my way though the code which writes into the `PersistedGrants` table I found these bits of code which seem to be wrong.

The first parameter of the `GetConsentKey` function is the `subjectId` and the second one is the `clientId`. But all invocations of this method pass `clientId` as first and `subjectId` as second parameter. This seems wrong to me.

I guess this is not evil but at least not really clean.

https://github.com/IdentityServer/IdentityServer4/blob/74827bc4df39c2234d9490ba5ce032ce806ab475/src/IdentityServer4/Stores/Default/DefaultUserConsentStore.cs#L34-L48

Mardoxx:
We switch impl in GetConsentKey, and also switch param order for calling functions.
This should not break any existing persisted keys.

### DIFF
--- a/src/IdentityServer4/Stores/Default/DefaultUserConsentStore.cs
+++ b/src/IdentityServer4/Stores/Default/DefaultUserConsentStore.cs
@@ -33,7 +33,7 @@ namespace IdentityServer4.Stores
 
         private string GetConsentKey(string subjectId, string clientId)
         {
-            return subjectId + "|" + clientId;
+            return clientId + "|" + subjectId;
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


Mardoxx:
Was thinking... possibly more consistent to ship a breaking change to `IUserConsentStore` and switch all the params to `clientId`, `subjectId` ?

brockallen:
Yep, good catch. Thanks.

brockallen:
> Was thinking... possibly more consistent to ship a breaking change to IUserConsentStore and switch all the params to clientId, subjectId ?

On this... what do you mean? I don't think it really matters given the fix you submitted?

Mardoxx:
Thanks

I was having a brief look at other functions' signatures in idsrv and it seems to use clientId followed by subjectId.
But yes you are right, with this patch it doesn't matter :)

PR #1723, opened 2017-11-09
small clarification

Triwaters:
I'm assuming it's the text that doesn't quite match the proper image, and not the other way around.


brockallen:
Yep, and Thanks!

PR #1725, opened 2017-11-09
excise seemingly superfluous instructions

Triwaters:
I could be way off, and perhaps other users with other setups are seeing something different, but every time I play with the code/instructions in this section, I have never yet needed to manually add anything to the .csproj file as instructed. The Tools.DotNet section is always already there.


brockallen:
It depends on how you have created your project. If you started with empty, you will need to add this. If you started with MVC template you prolly won't need to do this step.

Triwaters:
That's ironic, as the other PR I suggested today was a clarification on project type. Following the image, as I have been, means getting the MVC template. Should we assume MVC? In which case (I'll nod and pretend I knew this) the manual additions won't be required. If we assume the empty template, we'd want to go back and re-work the text/image instructions on project setup.

brockallen:
I'd have to reacquaint myself with all the steps that we wrote into the quickstarts, and also some of these steps were hold overs from project.json, so it might be possible that this step is no longer needed. I'll have a look when I gets some time.

brockallen:
I added a blurb on this. I still think it's necessary,. especially if you are using the prior quickstart's "after" code as the starting point for this quick start.

PR #1729, opened 2017-11-09
lost a day on this

Triwaters:
The quickstart code and the tone of the dialog agree that at this point we are using data-driven logins, but this implementation in the documentation assumes hard-wired users. As a side-bar note, the failure to call .AddAspNetIdentity<ApplicationUser>() is one of the reasons users (yours truly being one) run into the "sub claim is missing" problem. All these assertions subject to verification, but this is what I'm seeing. :)

### DIFF
--- a/docs/quickstarts/8_entity_framework.rst
+++ b/docs/quickstarts/8_entity_framework.rst
@@ -69,7 +69,7 @@ We will replace them with this code::
     // configure identity server with in-memory stores, keys, clients and scopes
     services.AddIdentityServer()
         .AddDeveloperSigningCredential()
-        .AddTestUsers(Config.GetUsers())
+        .AddAspNetIdentity<ApplicationUser>()
         // this adds the config data from DB (clients, resources)
         .AddConfigurationStore(options =>
         {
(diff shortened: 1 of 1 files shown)


brockallen:
This sample is only meant to show EF in isolation, not using ASP.NET Identity. You can see this by the sample code: https://github.com/IdentityServer/IdentityServer4.Samples/blob/release/Quickstarts/8_EntityFrameworkStorage/src/QuickstartIdentityServer/Startup.cs#L26

Triwaters:
I see what you mean, but am wondering if other users/readers might be confused as I was by what was, to my eyes, the last lap--we're in the database now!--when in fact it is still hard-wired for in-memory users. I see that the next section mentions quick-starts including ASP.NET, but it wasn't until I read your comment just now that I revisited the pages and see what you're saying. Perhaps a paragraph highlighting this remaining step would help? I'd write it for you and suggest it, if you think that would be helpful. Thanks again for this tool and for your patience. I'm coming from ground zero on 90% of the technologies involved, so I lose a ton of time on vague things like these due chiefly to my ignorance.

brockallen:
The reason the sample is that way is because lots of people don't use ASP.NET Identity.

Feel free to change the PR with a note so others don't fall into the same assumption.

Triwaters:
will adjust the doc

PR #1730, opened 2017-11-09
Update standard IDistributedCache implementation.

dchristensen:
I was unable to find anything about `IDistributedInMemoryCache` anywhere, but MS provides `MemoryDistributedCache` as part of the `Microsoft.Extensions.Caching.Memory` package.


brockallen:
Yep, thanks.

PR #1731, opened 2017-11-10
added warning regarding static users list

Triwaters:
Per discussion with Brock, this proposed ending to the section alerts users to something they'll likely want to revise -- getting user management into ASP.NET Identity or some other tool


brockallen:
Looks, good, thanks.

Issue #1583, opened 2017-10-03
Try to detect non-unique scope names acrosss API and Id resources


brockallen:

brockallen:
Done

PR #1733, opened 2017-11-11
detect non-unique scope names #1583

brockallen:
#1583


src/IdentityServer4/Extensions/IResourceStoreExtensions.cs
```diff
@@ -25,6 +26,20 @@ public static class IResourceStoreExtensions
             var identity = await store.FindIdentityResourcesByScopeAsync(scopeNames);
             var apiResources = await store.FindApiResourcesByScopeAsync(scopeNames);
 
+            // attempt to detect invalid configuration. this is about the only place
+            // we can do this, since it's hard to vet the values in the store.
```
user1336: `get`

Issue #1564, opened 2017-09-26
Add redirect_uri parameter to ErrorUrl


MNF:
On error page, based on IdentityServerHost,  we want to add a link to  the client site, e.g. "You can return to your site".
We want to use `redirect_uri` parameter, but it's not available on HomeController.Error action.
I found that redirection to error page implemented in
IdentityServer4\Endpoints\Results\AuthorizeResult.cs
  ```
async Task RedirectToErrorPageAsync(HttpContext context)
{
…
   var errorUrl = _options.UserInteraction.ErrorUrl;
            var url = errorUrl.AddQueryString(_options.UserInteraction.ErrorIdParameter, message.Id);
            context.Response.RedirectToAbsoluteUrl(url);
}
```
I want to add to url RedirectUri
 `    url = url.AddQueryString("redirect_uri",  Response.RedirectUri);`
and read it in HomeController.Error action.
Is it a good idea, or you see some issues with it or can suggest alternatives?
If you think, that it is ok, will it be useful for other users? Do you want me to create pull request?

brockallen:
No, we don't pass it along because in an error state we don't know if the redirect uri is safe or not. For example, is the reason we're showing the error page because the redirect uri is invalid?

MNF:
If I will add validation that redirect_uri is in a list of allowed redirect uris for the client, will it be safe to use such uri? The link will appear for most of the errors except invalid redirect uri.

brockallen:
We can have a look to see if we think this is safe or not.

brockallen:
I submitted a PR for this. It needs scrutiny.

brockallen:
@leastprivilege i put it back on the 2.1 milestone for discussion so we can decide to include or punt.

brockallen:
Done

PR #1736, opened 2017-11-11
Add redirect_uri parameter to ErrorUrl #1564

brockallen:
#1564


brockallen:
This one we need to pair review, @leastprivilege

PR #1752, opened 2017-11-15
could be off...

Triwaters:
but these tweaks seem in the right direction? Might be safe to say I'm asking as much as telling, here...


brockallen:
Yea, that looks about right.

PR #1755, opened 2017-11-15
where angels fear to trod

Triwaters:
Some issues in this page, but you'll need to edit it to resolve them, as they include missing words only you can supply. I also tweaked a reference to Claims improperly marked up. Elsewhere you left claims unmarked up, so I went that route.

### DIFF
--- a/docs/reference/profileservice.rst
+++ b/docs/reference/profileservice.rst
@@ -23,7 +23,7 @@ ProfileDataRequestContext
 Models the request for user claims and is the vehicle to return those claims. It contains these properties:
 
 ``Subject``
-    The ``ClaimsPrincipal`` modeling the user. If the request The claims from the user's cooke will be in the ``ClaimsPrincipal``.
+    The ``ClaimsPrincipal`` modeling the user. If the request [I don't pretend to know what you meant to say about the request! (But you didn't say it... this fragment is followed by a new sentence or perhaps the "then" that follows an incomplete "if" --> ] The claims from the user's cookie will be in the ``ClaimsPrincipal``.
 ``Client``
     The ``Client`` for which the claims are being requested.
 ``RequestedClaimTypes``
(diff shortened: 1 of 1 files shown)


docs/reference/profileservice.rst
```diff
@@ -23,15 +23,15 @@ ProfileDataRequestContext
 Models the request for user claims and is the vehicle to return those claims. It contains these properties:
 
 ``Subject``
-    The ``ClaimsPrincipal`` modeling the user. If the request The claims from the user's cooke will be in the ``ClaimsPrincipal``.
+    The ``ClaimsPrincipal`` modeling the user. If the request [I don't pretend to know what you meant to say about the request! (But you didn't say it... this fragment is followed by a new sentence or perhaps the "then" that follows an incomplete "if" --> ] The claims from the user's cookie will be in the ``ClaimsPrincipal``.
```
MovGP0: the annotation in brackets should be removed

docs/reference/profileservice.rst
```diff
@@ -41,7 +41,7 @@ IsActiveContext
 Models the request to determine is the user is currently allowed to obtain tokens. It contains these properties:
 
 ``Subject``
-    The ``ClaimsPrincipal`` modeling the user. If the request The claims from the user's cooke will be in the ``ClaimsPrincipal``.
+    The ``ClaimsPrincipal`` modeling the user. If the request [I don't pretend to know what you meant to say about the request! (But you didn't say it... this fragment is followed by a new sentence or perhaps the "then" that follows an incomplete "if" --> ] The claims from the user's cookie will be in the ``ClaimsPrincipal``.
```
MovGP0: the annotation in brackets should be removed

docs/reference/profileservice.rst
```diff
@@ -23,15 +23,15 @@ ProfileDataRequestContext
...
 ``Client``
     The ``Client`` for which the claims are being requested.
 ``RequestedClaimTypes``
     The collection of claim types being requested.
 ``Caller``
     An identifier for the context in which the claims are being requested (e.g. an identity token, an access token, or the user info endpoint). The constant ``IdentityServerConstants.ProfileDataCallers`` contains the different constant values.
 ``IssuedClaims``
-    The list of ``Claim``s that will be returned. This is expected to be populated by the custom ``IProfileService`` implementation.
+    The list of claims that will be returned. This is expected to be populated by the custom ``IProfileService`` implementation.
```
MovGP0: this was intentionally the name of the class and should net be changed in this way

Triwaters: OK, but even so there is some formatting issue, which is why I took a swing at that... on the currently live page, the entire sentence comes across as "an object" -- colored font red with a button border around it, plus some tick marks that shouldn't be seen. So at a minimum, a formatting fix on that sentence. I see your point on the class, that having been said.

docs/reference/profileservice.rst
Triwaters: Agreed -- but what we really need is the corrected text desired. I will remove my comment, but that leaves the two spots in need of correction by someone else. That make sense?

Triwaters: Probably a better way to handle that one, but I'm a novice on git. :)

PR #1761, opened 2017-11-16
two errant sentences left unmodified

Triwaters:
I did this -->   ``Claim`` s      so as to get the Claim word formatted properly without the entire sentence getting treated like object-text etc.

Also the two sentences worded as such are left uncorrected: If the request The claims from the user's....


brockallen:
that looks good, thanks!

PR #1771, opened 2017-11-17
Updated extension grant docs 

osudude:


leastprivilege:
This is not true. There are multiple supported ways to send client secrets - and this is even extensible.

PR #1796, opened 2017-11-23
Update 6_aspnet_identity.rst

mattlaver:
Specify DbContext when creating the user database. This resolves the "More than one DbContext was found" when running the command.


brockallen:
Cool, thanks.

PR #1798, opened 2017-11-23
Allows overriding methods in default services

lacroixdavid1:


brockallen:
We'd also want to promote any member variables to be protected, like we do elsewhere: https://github.com/IdentityServer/IdentityServer4/blob/dev/src/IdentityServer4/Services/DefaultTokenService.cs#L35

Also, we need you to sign the CLA, please.

dfrunet:
@lacroixdavid1, class DefaultUserSession is still internal, so no external overrids allowed : (
Could you fix that as well, please?

lacroixdavid1:
@dfrunet done

brockallen:
Thanks!

PR #1813, opened 2017-11-29
Doc update: Fix Configure after quickstart ui download

AndersAbel:
Removed duplicate UseMvc. Added UseStaticFiles, without that the wwwroot assets won't be served.


leastprivilege:
thanks!

PR #1893, opened 2017-12-20
Add ClientName to LogutMessage and LogoutRequest

kevinlo:
Add ClientName to LogutMessage and LogoutRequest so the AccountServices can set it to the LoggedOutViewModel.ClientName and show the client name, instead of client id in the UI.

### DIFF
--- a/src/Host/Quickstart/Account/AccountService.cs
+++ b/src/Host/Quickstart/Account/AccountService.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Brock Allen & Dominick Baier. All rights reserved.
+// Copyright (c) Brock Allen & Dominick Baier. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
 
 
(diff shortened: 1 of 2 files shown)


brockallen:
Thanks for this. Mind adding a fallback to ClientId if ClientName is null/empty?

kevinlo:
@brockallen.   Do you mean at the AccountServices?   I thought the  ClientId is normally a non-human readable string and the client should always have a name.

I check the [spec](https://tools.ietf.org/html/draft-ietf-oauth-dyn-reg-20#section-2), it says the authorization server MAY display to client_id if client_name is omitted.

> client_name
      Human-readable name of the client to be presented to the user
      during authorization.  If omitted, the authorization server MAY
      display the raw "client_id" value to the user instead.  It is
      RECOMMENDED that clients always send this field. 

As it just say MAY and it RECOMMEDED that clients always send the client_name, I don't do it.

Also, the AccountServices in the QuickStart is provided as the sample codes only, right?   People can change it to what they want.

If you guys think it should fallback, I can put it as

```
            var vm = new LoggedOutViewModel
            {
                AutomaticRedirectAfterSignOut = AccountOptions.AutomaticRedirectAfterSignOut,
                PostLogoutRedirectUri = logout?.PostLogoutRedirectUri,
                ClientName = logout?.ClientName ?? logout?.ClientId,
                SignOutIframeUrl = logout?.SignOutIFrameUrl,
                LogoutId = logoutId
            };
```

What do you think?

brockallen:
Yes to all of the above :)

kevinlo:
??  null-coalescing operator does not check the string for empty.  I change it to:

`ClientName = string.IsNullOrEmpty(logout?.ClientName) ? logout?.ClientId : logout?.ClientName`

brockallen:
Looks good, thanks!

Issue #1946, opened 2018-01-05
Exception reading message: The provided payload cannot be decrypted because it was not protected with this protection provider.


johnrutherford:
### Issue

We upgraded from IdentityServer4 v1 to v2 about a month ago. Since then, we have been regularly seeing this error message in our logs. 

It doesn't appear to be causing any problems with our users, so we haven't been too concerned about it, but I would like to address it.

It looks like the error is being logged from the `ProtectedDataMessageStore` which was added in v2, but it's not immediately clear to me what was encrypting the messages in v1. 

### Relevant parts of the log file

```
Exception reading message: The provided payload cannot be decrypted because it was not protected with this protection provider.
```

brockallen:
Sounds like a data protection problem. Are you losing your data protection keys?

johnrutherford:
I looked at my data protection keys and it looks like a similar issue to this one:
https://github.com/aspnet/DataProtection/issues/187.

So it doesn't look like an IdentityServer issue. Thanks for pointing me in the right direction!

It would help if the full exception was logged, though, rather than just the exception message. I can do a pull request for that if you'd like.

leastprivilege:
yes please.

johnrutherford:
Ok, I'm not so sure that it is a DataProtection issue. That issue that I found was fixed in ASP.NET Core 2.0. 

My data protection keys are stored in a file share and I have the same app discriminator set for all load-balanced instances of the web app. I have verified that they are able to read the keys. I wrote a small console app that uses the same app discriminator and file share for keys. I was able to protect and unprotect messages between different machines with no issues. 

Any ideas of what else the issue could be?

OscarCanek:
The problem could be this:

https://github.com/dotnet/corefx/issues/22510

PR #1948, opened 2018-01-05
Log full exceptions when available to aid in debugging

johnrutherford:

This is related to issue #1946. Made changes to log full exceptions rather than just exception messages. This will aid in debugging since the full exception information, including inner exceptions will be available in logs.

No.

- [ ] <strike>Unit Tests for the changes have been added (for bug fixes / features)</strike> N/A

**Other information**:

I was initially just going to do this in the `ProtectedDataMessageStore` class, but I found several other places where a full exception could be logged rather than just the exception message.


leastprivilege:
thanks!

brockallen:
Thanks!

Issue #1842, opened 2017-12-08
Removing unsafe-inline for Edge raises issues


pierslawson:
The bug fix https://github.com/IdentityServer/IdentityServer4/issues/1748 seems sensible but does raise a couple of issues...

For me this broke the redirection from the authorize endpoint when I was using Edge. Somehow I still had a 2016 version of Edge installed, despite having run all available Windows Updates. I only managed to get it working by downloading the Windows Update Assistant from https://www.microsoft.com/en-gb/software-download/windows10 and forcing an install of the Windows 10 Fall Creators Update. So the first issue is... I wonder how many other "real" users might encounter the same problem.

Secondly, would it be sensible to include a submit button in the form rendered by the connect/authorize endpoint so that the form can be manually submitted if the user's browser doesn't support JavaScript (for whatever reason)? The button could be hidden by JavaScript (if it is allowed to execute) before the form is submitted. Or is there a security reason you didn't include a button on the page?

brockallen:
There's an open issue (or PR?) already for your second part.

brockallen:
Fixed the noscript part here: #1880.

brockallen:
As for the first part of your question -- I don't know what to say. Edge, I thought, was under the same automatic release model that Chrome and FF use, so in theory it should always be up to date?

brockallen:
@pierslawson any thoughts?

pierslawson:
I don't think it auto updates in the same way Chrome does. It appears to update as part of significant OS updates e.g. Creators Fall Update. There is plenty of chatter about Microsoft potentially breaking this link (e.g. https://www.neowin.net/news/microsoft-will-unbundle-its-edge-browser-from-the-windows-10-os-with-redstone-3) but I can't see anything confirming it. So updates seem to be optional and in some cases (such as mine) the update mechanism seems to get stalled. 

Since Edge appears to have supported the headers correctly for quite a while, maybe it is time to move forward. My only real issue was that it took quite some time to realise what the problem was... then even longer to force an update of Edge. Perhaps a temporary note in the documentation might help the discoverability of the work around (i.e. check your version of Edge and ensure it is greater than X then force an update by using Windows Update or the Windows Update Wizard or Chrome / Firefox).

brockallen:
> Perhaps a temporary note in the documentation might help the discoverability of the work around

I'm not sure where that would go. Any suggestions?

pierslawson:
[...]

PR #2035, opened 2018-01-29
Add  Content-Security-Policy options

dlemstra:
This pull requests adds two new options that can be used to control the Content-Security-Policy options.

- The option `Level` add backwards compatibility with browsers that only support CSP level 1.
- The option `AddDeprecatedHeader` can be used to disable the `X-Content-Security-Policy` header.

Nope, but I wonder if we should move `AuthenticationOptions.RequireCspFrameSrcForSignout` to the new `CspOptions` class.

**Other information**:
This pull request addresses the issue reported in #1842.


brockallen:
Hi @dlemstra. Sorry for the delay in looking at this. I like the idea and the PR. @leastprivilege and I will discuss this week to see if we want to get this into our 2.2 release. Thanks!

brockallen:
Ok, I took a look (and even ran/debugged thru it) and it looks good to me. The short summary is that if you set CSP2, we don't emit the 'unsafe-inline' to support old browsers, if you set CSP1 we do emit it so older browsers will work (but disabling CSP given their lack of support for CSP2's hash).

dlemstra:
@brockallen It was a while ago when I researched this but I think your summary is correct.

Issue #2039, opened 2018-01-31
 Bug in Basic Authentication parsing for client credentials


wallymathieu:
We have noticed that some of our (NodeJS based) clients have started having issues connecting to IdentityServer4 using existing code. There are two issues related to this on the library that we use:

https://github.com/lelylan/simple-oauth2/issues/148

https://github.com/lelylan/simple-oauth2/issues/167

As far as we have found, we can't use any special chars in client secrets in newer versions of the above lib. This kind of issue sounds a bit hairy that could end up breaking existing applications if behavior is changed.

DiginessForever:
Have you posted a question to Stack Overflow?  Please do and post the link here.  You'll likely need to post at least some code fragments from your clients to go with that question.

wallymathieu:
What kind of question would be suitable for stackoverflow? 

>> Clients in possession of a client password MAY use the HTTP Basic
>> authentication scheme as defined in [RFC2617] to authenticate with
>> the authorization server. The client identifier is encoded using the
>> "application/x-www-form-urlencoded" encoding algorithm per
>> Appendix B, and the encoded value is used as the username; the client
>> password is encoded using the same algorithm and used as the
>> password. - [Source](https://tools.ietf.org/html/rfc6749#section-2.3.1)

The code to reproduce it is simply to use the above mentioned library to connect to identity server with a password that is different when url encoded.

https://github.com/lelylan/simple-oauth2/blob/master/example/index.js
```
 const oauth2 = simpleOauthModule.create({
    client: {
      id: 'client.id',
      secret: 'asfdasd+fdsafdsa^fdsafsd&',
    },
    auth: {
      tokenHost: 'https://id.somedomain.com',
      tokenPath: '/connect/token',
      authorizePath: '/connect/authorize',
    },
  });
```

wallymathieu:
Since I've not read the spec, I was hoping that someone else could read it and say : yay or nay. If this is something that is wrong in identity server, then changing it might break things ...

leastprivilege:
yes 

https://identityserver4.readthedocs.io/en/release/intro/specs.html

wallymathieu:
thanks @leastprivilege

wallymathieu:
I'm closing the question, since it's answered.

wallymathieu:
I currently have not enough time to file a full bug report. But someone with more time could perhaps do that ...

arekkas:
[...]

PR #2052, opened 2018-02-06
Fix Basic Authentication encoding to fully comply with RFC6749

leastprivilege:
see #2039


leonardlabat:
Hello

Any chance for this to be also patched on IS3 ?
Will you accept a PR about that ?

Thanks

leastprivilege:
Not sure this is a good idea - this will break many existing systems. Upgrade to IS4.

PR #2091, opened 2018-02-20
Update unhandled exception logging

### LINKED ISSUE #2090: Use Exception overload of ILogger.LogCritical for unhandled exceptions
Unhandled exceptions are caught in `IdentityServerMiddleware` this way:

`_logger.LogCritical("Unhandled exception: {exception}", ex.ToString());`

Since this is not using the overload of `LogCritical` that takes an exception object, registered loggers do not receive the full exception details, making it difficult to aggregate resulting log data based on just the exception message. When using Kestrel, the exception is caught again and an event is emitted with the exception object. 

Consider something like:

`_logger.LogCritical(ex, "Unhandled exception: {exception}", ex.Message);`

chrisowhite:
* Use an overload of `LogCritical` that accepts an Exception object.

Resolves #2090


brockallen:
Excellent, thanks!

chrisowhite:
Any time! Thank you!

Issue #2098, opened 2018-02-22
Using HttpMethods.IsGet() and HttpMethods.IsPost() instead of string comparison


Icidis:
Hi,

I would like to suggest that all checking for HttpMethod verbs be done by using the static helpers methods provided in Microsoft.AspNetCore.Http.Abstractions.

**Changing from:**
`if (context.Request.Method != "GET")`

**To:**
`if (!HttpMethods.IsGet(context.Request.Method))`

Reference to Microsoft Docs:
[HttpMethods Class (Microsoft.AspNetCore.Http)](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.httpmethods?view=aspnetcore-2.0)

I would like to do a pull request for the changes but as per your contibution guidelines I'm opening this issue first.

Please let me know your thoughts.

leastprivilege:
Ok. Go ahead. Thanks!


-------
Dominick Baier

brockallen:
merged, thanks!

PR #2099, opened 2018-02-22
Use HttpMethods.IsGet() and HttpMethods.IsPost() instead of string comparison

Icidis:
Closes issue (#2098)


brockallen:
great, thanks!

PR #2109, opened 2018-02-26
Update authorize.rst Fix IdentityModel example

JensLeible:
Use RequestUrl class instead of nonexistent AuthorizeRequest

authorize.rst -> IdentityModel example states to use nonexistent AuthorizeRequest instead of RequestUrl (which I consider to the be right pick here?)

no


leastprivilege:
thanks!

PR #2121, opened 2018-03-01
explicitly set the default value of base target in html response from AuthorizeResult

garymacpherson:
Currently the authorization code flow with response method of form_post works unless attempting to do so from within a dialog window in IE. This is due to the html being returned from `IdentityServer4.Endpoints.Results.AuthorizeResult.cs` not explicitly setting the default value of base target to _self.

If not defined, all other browsers simply imply this value anyway, it purely affects dialog windows created from IE which due to a well known and publicized bug, requires it to be set explicitly.

All other html pages are fine as they can be controlled by the implementation, however this is the only html I have found that is served directly from Identity Server that is affected in this way.

No

- [*] The commit follows our [guidelines](https://github.com/IdentityServer/IdentityServer4/blob/dev/.github/CONTRIBUTING.md)
- [*] Unit Tests for the changes have been added (for bug fixes / features)

**Other information**:
Documentation of the default value https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base

### DIFF
--- a/src/IdentityServer4/Endpoints/Results/AuthorizeResult.cs
+++ b/src/IdentityServer4/Endpoints/Results/AuthorizeResult.cs
@@ -169,7 +169,7 @@ namespace IdentityServer4.Endpoints.Results
             return uri;
         }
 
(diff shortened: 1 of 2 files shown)


brockallen:
Thanks!

PR #2127, opened 2018-03-02
Fix invalid grant type validation result

### LINKED ISSUE #2116: Wrong GrantValidationResult error value
I get the wrong error value from GrantValidationResult inside my custom IResourceOwnerPasswordValidator.

Here is the code:

    if (roles.Contains(Constants.ROLES_CLIENT))
    {                
        if (!user.PhoneNumberConfirmed)
            {
                 context.Result = new GrantValidationResult(TokenRequestErrors.InvalidRequest, "Phone 
                 number not verified");
                 return;
            }
    }

When the token endpoint is called, I get the following:

    {
       "error": "invalid_grant",
       "error_description": "Phone number not verified"
    }

I try all types of TokenRequestErrors but I always get invalid_grant.

I also try to change this value using the following line but it also doesnt work: context.Result.Error = "whatever";

user1336:
Fixes #2116

No

- [ X] The commit follows our [guidelines](https://github.com/IdentityServer/IdentityServer4/blob/dev/.github/CONTRIBUTING.md)

**Other information**:


leastprivilege:
thanks!

Issue #2114, opened 2018-02-28
Add more configuration validation to token/authorize pipeline


leastprivilege:
initiated by https://github.com/IdentityServer/IdentityServer4/issues/2072

let's collect more config checks here and then implement them in one go.

* AccessTokenLifetime is 0

what else?

LindaLawton:
1. Browser enabled with no redirect URI set?    
2. Using JavaScript without setting cors 
3. RequireClientSecret missing a secret?  

I don't think its worth invaliding the max tokenlifetime,  being that its an int the potential age of the access token is greater then the epoch time.   I think you and I discussed this on stack. [link](https://stackoverflow.com/a/48149679/1841839)

brockallen:
Ideas:
 * maybe add a check to `IClientStoreExtensions.FindEnabledClientByIdAsync` since most (if not all) of our code should be funneling thru that API to load a client
* maybe add a new service in DI to do the validation. possibly devs could replace to add their own checks? On the values in `Properties` for example?

LindaLawton:
On the values in Properties for example?  <---- Love that idea since i am using properties.

chrisowhite:
Are you guys looking to implement specific default rules for 2.x or would that be a 3.0 thing (saw the comment from @leastprivilege in the above PR)?

leastprivilege:
I will merge the base infrastructure today. But it will be no-op by default. Then we can "crowd source" validation rules to be used optionally. They will then become the default in 3.0

leastprivilege:
OK - here's the new validator:

https://github.com/IdentityServer/IdentityServer4/blob/dev/src/IdentityServer4/Validation/DefaultClientConfigurationValidator.cs

It's not enabled by default - but in the host it is:
https://github.com/IdentityServer/IdentityServer4/blob/dev/src/Host/Startup.cs#L52

I added a bunch of validation checks already. Feel free to PR to add new ones!

Each check should have a corresponding test here:
https://github.com/IdentityServer/IdentityServer4/blob/dev/test/IdentityServer.UnitTests/Validation/ClientConfigurationValidation.cs

LindaLawton:
What happens if no context.Client.PostLogoutRedirectUris is supplied?

leastprivilege:
Well - that's an optional setting.

leastprivilege:
https://github.com/IdentityServer/IdentityServer4/issues/2317

janpieterz:
[...]

PR #2200, opened 2018-04-04
Add client configuration validation infrastructure

leastprivilege:
https://github.com/IdentityServer/IdentityServer4/issues/2114


leastprivilege:
OK - the base infrastructure is done but the validator will not do anything by default because it might break existing code.

PR #2239, opened 2018-04-17
Fix setting BearerToken

M0ns1gn0r:
The `SetBearerToken()` extension method is located in `IdentityModel.dll` which won't be referenced in the MvcClient if one follows the walk-through.


leastprivilege:
thanks!

PR #2298, opened 2018-05-04
Extension Grant flows need all the data of the request at the final build of the claims.

ghstahl:
Given an Extension Grant as an entry condition, there is a need for the IProfileService to have knowledge of the incoming request parameters.
At the moment the only way to pass that knowledge of incoming parameters was to make up a fake claim on the identity in the IExtenstionGrantValidator implementations.

Given that extension grants are custom, the need for all data to be known when the final claims are built should be allowed.

Probably, I added a new property to a class.  My C# chops on that type of addition is a little foggy right now.  If it does, I guess I could inherit and add an extension that upcasts to pull the new property.  My understanding is that adding a new property will not break unless some serialization/deserialization happens.  I can't imagine anyone doing that to a temporary context object.

Since this is a pass along parameter it is up to the IProfileService implementation to do unit tests on incoming ValidateRequest object in the context.

**Other information**:


ghstahl:
@leastprivilege No issue with the ctor param to a property change.  Do you want to do it, or have me do another PR?

leastprivilege:
Update the pr plz

leastprivilege:
thanks!

PR #2306, opened 2018-05-10
Add device flow authorize request validator

scottbrady91:


leastprivilege:
Looks good.

leastprivilege:
Tests?

scottbrady91:
What are they?

In the proof of concept, I have tests using Moq that I'm in the process of adapting to the same style as your existing tests. Unless you're happy for me just to bring them across?

leastprivilege:
Ah OK - I would prefer to keep the style consistent.

PR #2318, opened 2018-05-16
API example uses userinfo endpoint

LindaLawton:
When running though this sample it does not show anything about creating an api at `http://localhost:5001/identity`.   So unless you have an API set up with its IP and route you cant actually test this.   

By doing this you can automatically test that the bearer token is in fact working rather than having to set up an API on port 5001 which wasn't part of this sample anyway.

No

**Other information**:

I can add a new image for the one below 

>And click the “API” button to invoke the web API: 

If you guys want but I don't want to bother if your not interested in this change.

### DIFF
--- a/docs/quickstarts/7_javascript_client.rst
+++ b/docs/quickstarts/7_javascript_client.rst
@@ -172,7 +172,7 @@ Add this code to implement those three functions in our application::
 
     function api() {
         mgr.getUser().then(function (user) {
-            var url = "http://localhost:5001/identity";
+            var url = "http://localhost:5000/connect/userinfo";
 
             var xhr = new XMLHttpRequest();
             xhr.open("GET", url);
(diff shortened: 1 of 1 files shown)


brockallen:
DOesn't this example build on the prior examples which do? Also, oidc-client already/automatically has accessed userinfo for you, so showing the same calling it might add confusion.

brockallen:
> How about i change this to add a link to the sample on creating the API?

Sure. Also, I think at the top of each QS it says something like "we expect you to have done the prior QSs".

brockallen:
Ok, I think it was only on the first one:

> The quickstarts provide step by step instructions for various common IdentityServer scenarios. They start with the absolute basics and become more complex - it is recommended you do them in order.

LindaLawton:
There try that.   

Off topic:  What is this mark down called?

brockallen:
Ok, seems good.

The docs are built with sphinx: http://www.sphinx-doc.org/en/master/, and the markdown is RST: http://docutils.sourceforge.net/rst.html

PR #2320, opened 2018-05-17
Adds AllowAccessTokensViaBrowser to Implicit client example

LindaLawton:

The sample for creating an Implicit client should really contain ` AllowAccessTokensViaBrowser = true`

Nopes

**Other information**:

Copying doucmentation to anwser questions on SO FTW [identityserver4 with redux -oidc client requested access token - but client is not configured to receive access tokens via browser
](https://stackoverflow.com/a/50389408/1841839)

### DIFF
--- a/docs/quickstarts/3_interactive_login.rst
+++ b/docs/quickstarts/3_interactive_login.rst
@@ -217,7 +217,9 @@ Add the following to your clients configuration::
                 {
                     IdentityServerConstants.StandardScopes.OpenId,
                     IdentityServerConstants.StandardScopes.Profile
-                }
+                },
+
+                AllowAccessTokensViaBrowser = true
             }
         };
     }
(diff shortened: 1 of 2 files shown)


brockallen:
If your client app is a server-side app and it needs access tokens then it would use hybrid flow. I'd have to go back and read all the text in this QS to know if that's what this MVC client is trying to do.

LindaLawton:
This example is for Implicit which is used for [Adding a JavaScript client](http://docs.identityserver.io/en/release/quickstarts/7_javascript_client.html)

>This quickstart will show how to build a JavaScript client application. The user will login to IdentityServer, invoke the web API with an access token issued by IdentityServer, and logout of IdentityServer.

Being JavaScript I think its kind of going to require that you have `AllowAccessTokensViaBrowser `

Issue #2295, opened 2018-05-04
Client credentials flow with no scopes breaks after implicit flow set up


maximpashuk:
Using IdentityServer 2.1.3, but I think 2.2 also affected.

Imagine the situation:
1. Client set up client_credentials flow, client provide **no scopes** to get OAuth token for API access. By default it get all available scopes that set up in IdentityServer, and it is fine.
2. Additionaly client set up implicit flow for OpenIdConnect authentication with scopes "openid", "profile". Users of that client can successfully authenticate.

But BAM: first scenario with client_credentials starts failing with error **"cannot request OpenID scopes in client credentials flow"**.

If client specify scopes explicitly for first scenario, then all the things start work fine again.
But I think that **adding new grant type should not break previous grant types**.

Practically, I think then client specify no scope in token request, all available scopes for client should be filtered by grant_type.

P.S. I found issues like mine, for example in #632 and #1474, but only recommendation I found is to explicitly specify claims in client_credentials.

LindaLawton:
IIR the client_credentials flow is not meant to be used for server to server communication.  In this scenario there is no end user to identify. The client_credentials grant type is undefined in OpenID Connect.  

There for the open id scopes will not work for it as its not a real user.

leastprivilege:
IOW - in the client creds flow scenario with no scopes specified, we should filter out identity scopes... @brockallen

LindaLawton:
Verified

     var disco = await DiscoveryClient.GetAsync("http://localhost:5000");
     var tokenClient = new TokenClient(disco.TokenEndpoint, "ef2e55b8-dab8-4a7b-9f7e-5ce4351803ca.xena.biz", "1f0ec2de-3e57-4f64-abca-4f1ff725ab10");
      var fields = new Dictionary<string, string>
            {
                { "grant_type", "client_credentials" }
            };

        //fields.Add("scope", "testapi");
        var tokenRespone = await tokenClient.RequestAsync(fields, CancellationToken.None);

Results in:

> fail: IdentityServer4.Validation.TokenRequestValidator[0]
>       ef2e55b8-dab8-4a7b-9f7e-5ce4351803ca.xena.biz cannot request OpenID scopes in client credentials flow
> fail: IdentityServer4.Validation.TokenRequestValidator[0]
>       {
>         "ClientId": "ef2e55b8-dab8-4a7b-9f7e-5ce4351803ca.xena.biz",
>         "ClientName": "Client Credentials",
[...]

PR #2324, opened 2018-05-18
adds check for Client credentials and removes open id scopes

LindaLawton:

[2295 Client credentials flow with no scopes breaks after implicit flow set up](https://github.com/IdentityServer/IdentityServer4/issues/2295)

When you are using GrantTypes.ClientCredentials and not send any scopes.  IDs4 automatically grants access to all of the scopes currently set in client.AllowedScopes this is a problem if the open id (openid, profile) had been added.  As these scopes are not valid for the grant type client credentials. 

nope.

**Other information**:

I am happy to add a unit test for this if you like.  I just haven't checked out the test projects yet.  I added the open id scopes to the host client already.   I would almost think that any existing test we have would already be picking that up but i haven't checked.

-  Do we have constants for profile and openid?  I hate the idea of having them hard coded. 
-  Would this be better as an extension method?   `clientAllowedScopes.ValidateForGrantTYpe(grantType)
`

### DIFF
--- a/src/Host/Configuration/Clients.cs
+++ b/src/Host/Configuration/Clients.cs
@@ -26,7 +26,7 @@ namespace Host.Configuration
                     },
 
                     AllowedGrantTypes = GrantTypes.ClientCredentials,
-                    AllowedScopes = { "api1", "api2.read_only" }
+                    AllowedScopes = { "api1", "api2.read_only" , "openid", "profile" }
                 },
 
                 ///////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


LindaLawton:
The No_Scopes test is failing.   Because its not failing?

So i guess the question is do we need this test?  or should it be modified to be No_Scopes_WithOpenID?

I dont know enough about the testing plan here.   Feedback needed :)

TomCJones:
this is an oxymoron
from the introduction to OpenID Connect
OpenID Connect implements authentication as an extension to the OAuth 2.0 authorization process. Use of this extension is requested by Clients by including the openid scope value in the Authorization Request.

LindaLawton:
client_credentials flow is meant to be used for server to server communication. In this scenario there is no end user to identify. The client_credentials grant type is undefined in OpenID Connect. There for the open id scopes will not work for it as its not a real user.

leastprivilege:
I did a PR here https://github.com/IdentityServer/IdentityServer4/pull/2358

LindaLawton:
You kind of made a lot of extra work for yourself.   I would have happy made any changes you required.

PR #2336, opened 2018-05-27
Fix incorrect log message

shaikatz:
The LogInformation was printing the Idp twice instead of printing the currentIdp.

No

- [x] The commit follows our guidelines

### DIFF
--- a/src/IdentityServer4/ResponseHandling/AuthorizeInteractionResponseGenerator.cs
+++ b/src/IdentityServer4/ResponseHandling/AuthorizeInteractionResponseGenerator.cs
@@ -165,7 +165,7 @@ namespace IdentityServer4.ResponseHandling
             {
                 if (idp != currentIdp)
                 {
-                    Logger.LogInformation("Showing login: Current IdP ({idp}) is not the requested IdP ({idp})", currentIdp, idp);
+                    Logger.LogInformation("Showing login: Current IdP ({currentIdp}) is not the requested IdP ({idp})", currentIdp, idp);
                     return new InteractionResponse { IsLogin = true };
                 }
             }


src/IdentityServer4/ResponseHandling/AuthorizeInteractionResponseGenerator.cs
```diff
@@ -165,7 +165,7 @@ public class AuthorizeInteractionResponseGenerator : IAuthorizeInteractionRespon
             {
                 if (idp != currentIdp)
                 {
-                    Logger.LogInformation("Showing login: Current IdP ({idp}) is not the requested IdP ({idp})", currentIdp, idp);
+                    Logger.LogInformation("Showing login: Current IdP ({currentIdp}) is not the requested IdP ({idp})", currentIdp, idp);
```
LindaLawton: I suggest using [$ - string interpolation](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated) instead.   Might as well follow [Boy Scouts rules](https://medium.com/@biratkirat/step-8-the-boy-scout-rule-robert-c-martin-uncle-bob-9ac839778385) here.

    Logger.LogInformation($"Showing login: Current IdP ({currentIdp}) is not the requested IdP ({idp})");

well spotted.

leastprivilege: String interpolation is not allowed in logging messages - can break the sink by injecting curly braces. We had this problem before...

LindaLawton: Really?   I have never had that issue.  Something specific to the logger your using here?  I will keep an eye out in the future thanks for the tip.

leastprivilege: > Really? I have never had that issue.

Because you never tried it ;)

LindaLawton: yup i have i always use String interpolation, and its in all my logging both APIs and in the identity server project. 

    _logger.LogWarning($"User ({user.Id}) LastFiscalId (not set)");

Thats in ProfileService.   

I think you would be surprised what i have tried ;o)

leastprivilege: Then you should read the docs on how ASP.NET Core logging actually works ;)

leastprivilege: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-2.0&tabs=aspnetcore2x

LindaLawton: On it  :)  This has started a great conversation here.

leastprivilege:
thanks!

PR #2355, opened 2018-06-04
Change sdk version from 2 to 2.x

LindaLawton:

no

no

**Other information**:

I will have a look at the tutorials but I think the sample project would also need to be updated if we do that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ If you encounter issues or find bugs, please open an issue in this repo here fir
 ## How to build
 IdentityServer is built against the latest ASP.NET Core 2.
 
-* [Install](https://www.microsoft.com/net/download/core#/current) the latest .NET Core 2 SDK
+* [Install](https://www.microsoft.com/net/download/core#/current) the latest .NET Core 2.1 SDK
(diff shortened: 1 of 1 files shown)


LindaLawton:
Ha what?

@leastprivilege you merged the 2.1 change in right?   That built fine didnt it?

leastprivilege:
The Linux build is not working yet...

LindaLawton:
OK as long as its not something i broke :)

leastprivilege:
How about 2.x ?

LindaLawton:
Yeah I was thinking the same but if we need to have a specific version of the SDK installed to run it it may be best to hard code it.  Right now we are on 2.1,  but if they release 2.2 will it work out of the box with 2.2 probably not until we upgrade to that.   It definitely wont work if you have 2.0 on your machine.  Which is technically .x

But yes this means we will need to update this file every time we change.   Which is not to much of a problem i am good at remembering to do random stuff like this.

You decide.

leastprivilege:
The dev branch will always be on the latest 2.x SDK

LindaLawton:
Updated

leastprivilege:
thanks!

PR #2357, opened 2018-06-04
Dont log SecurityTokenExpiredException as error, since it is not

jepperc:
When having "not smart" gateway that is checking accesstokens against the introspect endpoint, lots of errors are logged with SecurityTokenExpiredException, just because of expired tokens. My expectation would not be any error.

No, since only logging level is changed..

No tests of logging..

**Other information**:
It seems possible to also change the return value to "ExpiredToken" instead of "InvalidToken", but for the sake of simplicity and no relationship to my original problem, i kept it out for now. Any thoughts?

### DIFF
--- a/src/IdentityServer4/Validation/TokenValidator.cs
+++ b/src/IdentityServer4/Validation/TokenValidator.cs
@@ -314,6 +314,11 @@ namespace IdentityServer4.Validation
                     Jwt = jwt
                 };
             }
+            catch (SecurityTokenExpiredException expiredException)
+            {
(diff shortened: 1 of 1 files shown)


leastprivilege:
Why are you sending JWTs to the introspection endpoint?

leastprivilege:
still fine for me - thanks!

jepperc:
So the reasoning I can't tell.. it's way before my time, and something with a gateway and access token checking :) 
We are thinking about moving away though, since it doesn't seem like we will go to reference tokens.
Thank you!

PR #2358, opened 2018-06-05
Filter identity scopes and offline_access when no explicit scopes are specificed in client credentials

leastprivilege:
see https://github.com/IdentityServer/IdentityServer4/issues/2295

### DIFF
--- a/src/IdentityServer4/Validation/TokenRequestValidator.cs
+++ b/src/IdentityServer4/Validation/TokenRequestValidator.cs
@@ -329,7 +329,7 @@ namespace IdentityServer4.Validation
             /////////////////////////////////////////////
             // check if client is allowed to request scopes
             /////////////////////////////////////////////
-            if (!await ValidateRequestedScopesAsync(parameters))
+            if (!await ValidateRequestedScopesAsync(parameters, ignoreImplicitIdentityScopes: true, ignoreImplicitOfflineAccess: true))
             {
                 return Invalid(OidcConstants.TokenErrors.InvalidScope);
             }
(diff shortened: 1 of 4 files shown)


src/IdentityServer4/Validation/TokenRequestValidator.cs
```diff
@@ -573,21 +573,31 @@ public TokenRequestValidator(IdentityServerOptions options, IAuthorizationCodeSt
...
                 _logger.LogTrace("Client provided no scopes - checking allowed scopes list");
 
                 if (!_validatedRequest.Client.AllowedScopes.IsNullOrEmpty())
                 {
                     var clientAllowedScopes = new List<string>(_validatedRequest.Client.AllowedScopes);
-                    if (_validatedRequest.Client.AllowOfflineAccess)
+                    if (!ignoreImplicitOfflineAccess)
                     {
-                        clientAllowedScopes.Add(IdentityServerConstants.StandardScopes.OfflineAccess);
+                        if (_validatedRequest.Client.AllowOfflineAccess)
```
LindaLawton: You are sure that you dont want to log that you are removing them?

If a developer added them without knowing that they are not valid they may wonder why they are not there and there is no error.

leastprivilege: feel free to do a PR - log at debug level

leastprivilege: PR?

LindaLawton: Probably not I have some tasks at work that need to be completed before I go on vacation.    As its your code you probably know exact where to put it anyway.

Issue #2267, opened 2018-04-25
BasicAuthenticationSecretParser does not handle empty client_secret


TheBels:
We are on ID4 2.1.2 on Net Core 2.x. We also use Swashbuckle.AspNetCore for swagger documentation. Our setup has only a client_id specified - we are not using client_secret. This all worked fine with Swashbuckle 1.x.

The issue is that I'm upgrading to Swashbuckle 2.4.0 and the client_id and client_secret are now being passed via Authorization Basic header of the form `client_id:client_secret`. This also is fine.. except we have an empty client_secret. Thus the encoded string given to ID4 to parse is `client_id:` And when going into `BasicAuthenticationSecretParser.ParseAsync` it fails when checking for non-empty and non-null client_id and secret fields: https://github.com/IdentityServer/IdentityServer4/blob/becb2206826b1039687b3ec895fad5e1b26ef0c5/src/IdentityServer4/Validation/BasicAuthenticationSecretParser.cs#L97

### Issue / Steps to reproduce the problem

To reproduce: via Postman or similar you may confirm logging in with only x-www-form-urlencoded works fine, e.g.:

![login-success-with-x-www](https://user-images.githubusercontent.com/12475792/39274029-a936469e-48ae-11e8-8b71-2b1453c39305.png)

Further you may confirm that authentication works fine with passing in a dummy secret:

![login-success-with-dummy-secret](https://user-images.githubusercontent.com/12475792/39274044-ba2fb192-48ae-11e8-920f-a7a4e7d17918.png)

However, when omitting the dummy secret, that is encoding the string `my_client_id:`, then a 400 is returned due to the above-mentioned behavior in BasicAuthenticationSecretParser:

![login-fail-with-empty-secret](https://user-images.githubusercontent.com/12475792/39274105-ec52ea2c-48ae-11e8-8773-bbb23c5c92f9.png)

Thus it appears `BasicAuthenticationSecretParser` has a bug with clients without secrets. I searched around and didn't find a SO post nor an Issue raised so perhaps I'm missing something. Anyone have any hints/ideas on what I'm missing here?

leastprivilege:
Looks like it was by design. We can change that.

Do you want to do a PR including test?

TheBels:
Hi @leastprivilege  - thanks for confirming the behavior.

I don't mind taking a look but I wouldn't be able to get to it immediately. But we can certainly keep the ticket open and I (or others if desired) can come back to this.

[...]

PR #2374, opened 2018-06-09
Make client secret optional while parsing basic authentication secret

AliBazzi:
This PR add feature request in #2267 ticket

It doesn't introduce breaking change.

### DIFF
--- a/src/IdentityServer4/Validation/BasicAuthenticationSecretParser.cs
+++ b/src/IdentityServer4/Validation/BasicAuthenticationSecretParser.cs
@@ -94,10 +94,10 @@ namespace IdentityServer4.Validation
             var clientId = pair.Substring(0, ix);
             var secret = pair.Substring(ix + 1);
 
-            if (clientId.IsPresent() && secret.IsPresent())
+            if (clientId.IsPresent())
             {
                 if (clientId.Length > _options.InputLengthRestrictions.ClientId ||
-                    secret.Length > _options.InputLengthRestrictions.ClientSecret)
+                    (secret.IsPresent() && secret.Length > _options.InputLengthRestrictions.ClientSecret))
                 {
                     _logger.LogWarning("Client ID or secret exceeds allowed length.");
                     return notfound;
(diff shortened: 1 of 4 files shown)


AliBazzi:
Sorry didn't get it, can you elaborate more ?

brockallen:
> Sorry didn't get it, can you elaborate more ?

Well, look at the status:

> Some checks were not successful
> 1 failing and 1 successful checks
> continuous-integration/appveyor/pr — AppVeyor build failed

AliBazzi:
Oh yes, I noticed that, but it looks like it's related to Ubuntu where it failed to compile the integration test project, the windows image with visual studio passed all tests

I don't think the modified unit test broke that project

brockallen:
Ah yes, you're right -- I hadn't followed the link to know it was the unbutu build. I guess something environmental.

AliBazzi:
This is my second PR to Identity server, and both of them were very minimal changes, but my bad luck jumped in the verification process in both PR

brockallen:
No worries. We'll have a look. Thanks

AliBazzi:
Welcome :)

leastprivilege:
Thanks!

Could you also add an integration test - basically a client with empty secret calling into the token endpoint - I want to make sure, nothing breaks later in the pipeline because of a null secret.

AliBazzi:
Sure, working on it :)

AliBazzi:
@leastprivilege please take a look at the integration test when you have time.

leastprivilege:
You also need a test for an empty secret I guess - not just RequireClientSecret = false.

leastprivilege:
Add a client with require secret = true but an empty secret - see if that works as expected.

AliBazzi:
Hi @leastprivilege 

After investigating the code, I think the semantic of empty secret is not supported currently, what is  supported is a client that has RequireClientSecret set to `false`.
When that property is set to `false` it was breaking client id and empty secrets passed in the Basic Authorization header, this PR aimed to fix this bug in `BasicAuthenticationSecretParser`, so adding a test to verify that a client secret empty + RequireClientSecret set to `True` will not pass due to current implementation.

Please advice on the next step and correct me please if I'm wrong.

leastprivilege:
thanks!

PR #2418, opened 2018-06-27
Code refactoring on UpdateRefreshTokenAsync

cguldogan:
With this PR I would like to refactor and simplify Update RefreshToken Async method.

No

**Other information**:

### DIFF
--- a/src/IdentityServer4/Services/DefaultRefreshTokenService.cs
+++ b/src/IdentityServer4/Services/DefaultRefreshTokenService.cs
@@ -99,55 +99,71 @@ namespace IdentityServer4.Services
 
             if (client.RefreshTokenUsage == TokenUsage.OneTimeOnly)
             {
-                _logger.LogDebug("Token usage is one-time only. Generating new handle");
-
-                // delete old one
-                await RefreshTokenStore.RemoveRefreshTokenAsync(handle);
-
-                // create new one
-                needsCreate = true;
+                needsCreate = await RemoveRefreshToken(handle);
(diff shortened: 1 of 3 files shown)


src/IdentityServer4/Services/DefaultRefreshTokenService.cs
```diff
@@ -99,55 +99,71 @@ public DefaultRefreshTokenService(ISystemClock clock, IRefreshTokenStore refresh
...
-                }
-
-                refreshToken.Lifetime = newLifetime;
-                needsUpdate = true;
-            }
-
-            if (needsCreate)
-            {
+                needsCreate = await RemoveRefreshToken(handle);
                 handle = await RefreshTokenStore.StoreRefreshTokenAsync(refreshToken);
```
veikkoeeva: Should all of these `async`  tasks also have `.ConfigureAwait(false)`? It doesn't appear the codebase has much them, but maybe it's OK to add them as one refactor anyhow. This doesn't only help with "GUI systems", but also with systems like https://github.com/dotnet/orleans (maybe even ASP.NET Core).

leastprivilege:
Thanks. But since you changed an interface . this is a breaking change and has to wait til 3.0

cguldogan:
Hi, @leastprivilege, what shall do I now ?

src/IdentityServer4/Services/DefaultRefreshTokenService.cs
cornem: `ConfigureAwait(false)` is not necessary in ASP.NET Core, as it [doesn't have a synchronization context](https://stackoverflow.com/a/42058544/1746803).

PR #2422, opened 2018-06-28
Device Flow Response Validator

scottbrady91:
With `slow_down` funcitonality using IDistributedCache.

Some hardcoded strings for token errors waiting on IdentityModel release.

### DIFF
--- a/src/IdentityServer4/ResponseHandling/TokenResponseGenerator.cs
+++ b/src/IdentityServer4/ResponseHandling/TokenResponseGenerator.cs
@@ -88,6 +88,8 @@ namespace IdentityServer4.ResponseHandling
                     return await ProcessAuthorizationCodeRequestAsync(request);
                 case OidcConstants.GrantTypes.RefreshToken:
                     return await ProcessRefreshTokenRequestAsync(request);
+                case OidcConstants.GrantTypes.DeviceCode:
+                    return await ProcessDeviceCodeRequestAsync(request);
                 default:
                     return await ProcessExtensionGrantRequestAsync(request);
             }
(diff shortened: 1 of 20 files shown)


src/IdentityServer4/ResponseHandling/TokenResponseGenerator.cs
```diff
@@ -227,6 +229,68 @@ protected virtual Task<TokenResponse> ProcessPasswordRequestAsync(TokenRequestVa
...
+        /// <param name="request">The request.</param>
+        /// <returns></returns>
+        protected virtual async Task<TokenResponse> ProcessDeviceCodeRequestAsync(TokenRequestValidationResult request)
+        {
+            Logger.LogTrace("Creating response for device code request");
+
+            //////////////////////////
+            // access token
+            /////////////////////////
+            var (accessToken, refreshToken) = await CreateAccessTokenAsync(request.ValidatedRequest);
```
veikkoeeva: I was looking through the PRs and noticed `.ConfigureAwait(false)` isn't used in the codebase... But I wonder if I comment here in some PR and ask would it be OK to introduce them little-by-little? :)

veikkoeeva: Good point. There isn't one in ASP.NET Core and it appears they aren't writing that out, see at https://blog.stephencleary.com/2017/03/aspnetcore-synchronization-context.html for more authoritative explanation. (That article recommends it for libraries, though.)

This brings a question, how much of this code could be run elsewhere? As for an example, some [Orleans](https://github.com/dotnet/orleans/) people are experimenting with IS. The system works so that it has roughly one threas per core and it gives "single-thread guarantee" (https://dotnet.github.io/orleans/Documentation/Advanced-Concepts/External-Tasks-and-Grains.html), which could be problematic. (`Task.Run` can be used to excape blocking behaviour if needed, though.)

That being written, I don't know if this has been a concious decision as in to save in trouble of writing it everywhere or if it might avoid some other trouble. So I thought just to ask it in some context. :)

leastprivilege:
should I merge?

src/IdentityServer4/ResponseHandling/TokenResponseGenerator.cs
scottbrady91: Okay, I'd have to defer to Dom & Brock about support throughout the rest of the library From what I've seen `configureawait` is only necessary when targetting .NET Framework. 

I'm going to assume that this PR won't be the place to start any change in behavior though. Maybe you could open a separate issue?

scottbrady91:
Yes, please

PR #2499, opened 2018-07-31
fix ??-operator priority

Backs:
No issue

No

- [ x] Unit Tests for the changes have been added (for bug fixes / features)

**Other information**:


brockallen:
@leastprivilege maybe we can have a look at this monday. i don't know what the side effect of this change might be.

brockallen:
> What issue does this PR address?
> No issue

Can you explain why the change is right or better? Thanks.

Backs:
@brockallen problem is that `GetHashCode` will return `0` if `Value` or `Type` is null. So
`new Secret("a").GetHashCode() == new Secret("b").GetHashCode()`
And it's not the behavior you expect, I think.

leastprivilege:
I will have to merge manually since this file has moved.

Issue #2434, opened 2018-07-02
Consider redirect uri scheme blocked list


brockallen:
Not that lists of known bad things is ideal, but it won't hurt. Schemes to consider:

javascript:
file:
data:
mailto:
ftp:
blob:
about:
ssh:
tel:
view-source:
ws:
wss:

LindaLawton:
I am kind of think its more the responsibility of the owner of the server to ensure that only valid redirect URIs are created.   What are you planing on doing checking it when the client is verified?  Seams like a lot of unnecessary checking to me.  

_I added a check in my developer console app so that third party developer cant add invalid redirect uris._

brockallen:
> I am kind of think its more the responsibility of the owner of the server to ensure that only valid redirect URIs are created.

Yea, that doesn't always happen :)

brockallen:
merged

PR #2506, opened 2018-08-02
add invalid uri scheme validation

brockallen:
PR for #2434


brockallen:
have a look @leastprivilege. i wasn't sure where else to add this. i put it into the strict uri validator so that in theory if someone really didn't want it, then they could just not use the base class or re-implement from scratch. but if they do derive from it and use the base methods, then they'd still get the benefit.

Issue #2532, opened 2018-08-08
Consider resolving login url, et al from named options


brockallen:
So that LoginPath doesn't need to be explicitly set.

brockallen:
I looked into this @leastprivilege -- turns out we do the opposite; we set the handler options to use our settings. I now recall when we were designing this -- that was deliberate. We made the decision to make our settings the primary value, and we overwrote the settings on the cookie option. So now... we're changing our thinking?

brockallen:
I tested the above approach and it seems to work well. The only possible breaking behavior is if someone were previously setting the cookie options path to something invalid and then those were getting corrected by our default paths to "/Account/Login" for example. Otherwise, this seems like it's going to work out alright.

brockallen:
PR submitted @leastprivilege

plachor:
@brockallen this still might be a breaking change as although domain should be case insensitive rest of the request might not. So now instead of /account/login we get /Account/Login as you mentioned.

PR #2540, opened 2018-08-09
resolve login/logout url, et al from named options

brockallen:
#2532


brockallen:
yes, @leastprivilege  and i will try to have a look tomorrow if he's not already drinking beer that early in the morning. if he's not, i'd like to update the preview we put out today. :)

brockallen:
Turns out he was not even in the office today, so this will wait until prevew2, unfortunately. That's in September.

brockallen:
@alexdresko -- We're pushing a preview1-update1 right now, so you should be able to test with that.

Ciwan1859:
@brockallen does this pull request fix the issue with the Register and Login buttons not working when following [this guide](http://docs.identityserver.io/en/release/quickstarts/6_aspnet_identity.html)

brockallen:
@Ciwan1859 I don't know what issues you are referring to. If there's not an open issue for it, then open one please.

PR #2545, opened 2018-08-13
Make some internal types public to facilitate custom service implementations

thirkcircus:
We have had to create a few custom services for IdSvr4 interfaces.  In several cases we've had to copy/paste internal constants, extension methods or validation methods to facilitate reuse of the default services.

This PR simply makes a few types/members public so we can reference/reuse them from our custom service types.

I used a `#pragma` to disable warnings about no xml documentation for all the many constants members that are now public - I figure the constants are fairly self-explanatory.

I realise there may be reasons for having made these types internal but if not, it would be helpful to enable their reuse.

cheers

### DIFF
--- a/src/Constants.cs
+++ b/src/Constants.cs
@@ -14,7 +14,6 @@ namespace IdentityServer4
         public const string IdentityServerName               = "IdentityServer4";
         public const string IdentityServerAuthenticationType = IdentityServerName;
         public const string ExternalAuthenticationMethod     = "external";
-        public const string AccessTokenAudience              = "{0}resources";
         public const string DefaultHashAlgorithm             = "SHA256";
 
         public static readonly TimeSpan DefaultCookieTimeSpan = TimeSpan.FromHours(10);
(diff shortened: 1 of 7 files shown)


brockallen:
Hey Justin -- do you know which constants specifically? We have `IdentityServerConstants` already that's meant to be public, so what we'd want to do for this PR is move the ones to be made public into the designated class.

thirkcircus:
Hi - aha, i see!

I'm really just after `AccessTokenAudience`.  😄
 
I've just altered the PR (force-push) - have dropped the commit to publicise `Constants` and moved just the one member over to `IdentityServerConstants`.

thirkcircus:
Oh.  😐   
I've just gone to create a custom DiscoveryResponseGenerator and see that all of `Constants.ProtocolRoutePaths` is internal so I can't reuse it - any objection to my also moving that into public constants?

thirkcircus:
and `SigningAlgorithms` also.

leastprivilege:
...and why?

thirkcircus:
Why what?  Why are we creating custom implementations of idsvr4 interfaces?

For backwards compatibility (url encoding of client ids and secrets in idsvr4 broke some of our clients and some of our clients couldn't handle the change to always render scope claims in an array even when only one present) and integration reasons.  For eg we host on two origins with a single issuer.  I'm wanting to create custom discovery and override just the authorization endpoint to hardcode the origin for when users log in.  This would be as simple as copy/paste the DiscoveryResponseGenerator and alter one line - except the constants used by that type are internal so I have to copy/paste a whole bunch of them as well.

PR #2570, opened 2018-08-23
6_aspnet_identity.rst

TPullen21:
Incorrectly states "which replaces the call to UseIdentity" instead of "which replaces the call to UseAuthentication".


brockallen:
yep, good catch. thanks!

PR #2617, opened 2018-09-12
Change: error code in TokenValidator class

slotakonrad:
In TokenValidator.cs, when catching SecurityTokenExpiredException, method return InvalidToken code instead ExpiredToken. This change fixes this issue.

no

- [- ] Unit Tests for the changes have been added (for bug fixes / features)

**Other information**:

### DIFF
--- a/src/Validation/Default/TokenValidator.cs
+++ b/src/Validation/Default/TokenValidator.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Brock Allen & Dominick Baier. All rights reserved.
+// Copyright (c) Brock Allen & Dominick Baier. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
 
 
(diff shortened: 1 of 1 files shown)


slotakonrad:
> Do you have a specific reason for needing this error code?

Yes, in my implementation, this errorCode is important to complete correct accessToken validation.
Now, program recognizes token as invalid, which makes difficult further programming application behavior.
If error code is  "ExpiredToken", application knows that token need to be refreshed and not recognize it as invalid.

leastprivilege:
> Yes, in my implementation, this errorCode is important

How do you call the validator? or from where?

slotakonrad:
> > Yes, in my implementation, this errorCode is important
> 
> How do you call the validator? or from where?

When checking if access token is generated by me (using my certificate installed in my identityServer WebApi)

leastprivilege:
I found some time to track this down. In principal I am OK with this change - but we are "leaking" the error in the UserInfo endpoint. Whereas RFC6750 says this is the right format

> HTTP/1.1 401 Unauthorized
     WWW-Authenticate: Bearer realm="example",
                       error="invalid_token",
                       error_description="The access token expired"

This could be fixed in the `Error` method in `UserInfoEndpoint`.

I can do this - just wanted to document it.

leastprivilege:
OK - I added a check https://github.com/IdentityServer/IdentityServer4/commit/023d7844e86d6f055aabf04fb7f233c70f896956

leastprivilege:
thanks!

PR #2658, opened 2018-09-29
Corrected internal value for ParsedSecretTypes.JwtBearer

scottbrady91:
ParsedSecretTypes.JwtBearer is used for [JWT Bearer client authentication](https://tools.ietf.org/html/rfc7523#section-2.2).

Tests and secret parsing all use the correct value of `urn:ietf:params:oauth:client-assertion-type:jwt-bearer` but `ParsedSecretTypes.JwtBearer`, the internal value for storing secret type, uses `urn:ietf:params:oauth:grant-type:jwt-bearer`.

This is a breaking change for anyone using the old value explicitly in their code.


brockallen:
Thanks there fella!

PR #2695, opened 2018-10-08
Changed level from error to warn on refresh token

alnimin:
Expired tokens
Invalid tokens

Could we please lower the log level for refresh token errors? This would reduce log clutter caused by expired refresh tokens and race conditions with one-time-only refresh tokens. Same thing was done for Identity Server 3 by John Korsnes, commit b8a375bd of 09.02.2015


brockallen:
Yea, sounds ok. Thanks.

PR #2745, opened 2018-10-24
Fix object logging

shaikatz:
Objects being serialized as strings instead of sending pure objects with @ suffix to Serilog.
It makes it impossible to filter/enrich in serilog, and to analyze in the logging systems.

No

**Other information**:
I've tested it locally, before the change those logs arrived to the Serilog enricher as strings in the Properties.
After the change, the Serilog enricher receives real structed data that we can work with.


leastprivilege:
How does this behave with non-Serilog logging systems?

leastprivilege:
IOW - is the @ decomposition also a feature of plain ASP.NET Core logging?

shaikatz:
I still can't find a specific documentation about @ in standard .NETCore logging, 
But I ran a test locally without Serilog and with this line of code:
`Logger.LogInformation(nameof(ValidatedAuthorizeRequest) + Environment.NewLine + "{@validationDetails}", details);`

This is the output:
```info: IdentityServer4.Endpoints.AuthorizeEndpoint[0]
      ValidatedAuthorizeRequest
      {
        "ClientId": "SomeClientId",
        "ClientName": "SomeClientName",
        "RedirectUri": "https://some.redirect.com/auth",
        "AllowedRedirectUris": [
          "http://localhost:3000/auth",
        ],
        "SubjectId": "anonymous",
        "ResponseType": "id_token token",
        "ResponseMode": "fragment",
        "GrantType": "implicit",
        "RequestedScopes": "openid",
        "State": "4a6413622e6144179b80f606a679118e",
        "Nonce": "2008bd7b234849c1aa703ee9bdaaadf9",
        "AuthenticationContextReferenceClasses": [
          "idp:Google"
        ],
        "Raw": {
          "client_id": "SomeClientId",
          "redirect_uri": "https://some.redirect.com/auth",
          "response_type": "token id_token",
          "scope": "openid",
          "state": "4a6413622e6144179b80f606a679118e",
          "nonce": "2008bd7b234849c1aa703ee9bdaaadf9",
          "acr_values": "idp:Google"
        }
      }
```

IMO it looks good even without Serilog.

filipw:
This is a very good idea, however it's worth mentioning that this is sort of a quiet breaking change. 
If you forwarded these Identity Server logs into some log index before, the objects were simply strings, however now they would get destructured into objects, causing a type mismatch. This can lead to all kinds of mapping exceptions or lost logs in i.e. logstash / elastic search, unless you manually resolve that or rollover the index. While not a huge deal, I encountered that when upgrading to Identity Server 2.3.

leastprivilege:
yes.. you are right.

PR #2777, opened 2018-11-04
unnecessary boolean literals

milbrandt:
Avoid unnecessary boolean literals `true` and `false` in comparisons (ie if there is no nullable involved).
Shorter Code, IMHO often more readable like `!valid` instead of `valid == false`

No.

**Other information**:
Got complains from Sonarqube in the Quickstart part

### DIFF
--- a/host/Quickstart/Account/AccountController.cs
+++ b/host/Quickstart/Account/AccountController.cs
@@ -84,7 +84,7 @@ namespace IdentityServer4.Quickstart.UI
             {
                 if (context != null)
                 {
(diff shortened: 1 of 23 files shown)


brockallen:
> IMHO often more readable like !valid instead of valid == false

This sort of shows that it's an opinion.

>  Got complains from Sonarqube in the Quickstart part

No idea what that is.

Can you explain what a real benefit from this PR is? Given that this doesn't really change any logic, this seems like a purely stylistic PR.

Issue #1999, opened 2018-01-17
How to disable url redirect to LoginUrl while not logged in? And return 401 unauthorized status instead


mokeyish:
I configured UserInteractionOptions`UserInteractionOptions.LoginUrl=null`,But it cause an validating error.

leastprivilege:
Why do you want to do that?

mokeyish:
@leastprivilege My API server and IdentityServer has runing together，The font-end  I have used Angular already,dosen't need to redirect LoginUrl

leastprivilege:
so why would you then do a request to the authorize endoint?

mokeyish:
@leastprivilege I have  solved it.
![image](https://user-images.githubusercontent.com/16131917/35075988-0ffc0adc-fc31-11e7-82cc-395c384b112c.png)

PR #2808, opened 2018-11-17
Support hosting the identity server as a WebAPI

paulcsiki:
If you host the identity server on a WebAPI project and you authenticate using a SPA framework like Angular you want to be able to setup an interceptor to detect if the user is no longer authenticated (cookie expired, never authenticated, etc) by reading the status code of XHRs. In the current implementation the code tries to perform a redirect which is not the ideal case for a WebAPI implementation.

No.

- [X] Unit Tests for the changes have been added (for bug fixes / features)


paulcsiki:
This PR resolves issue #1999.

paulcsiki:
This needs to be properly implemented by adding a configuration flag to indicate that IdentityServer4 should operate in a WebAPI mode and to never return redirects or write to the Response HTML.

PR #2856, opened 2018-11-30
Update Quickstarts

leastprivilege:


docs/quickstarts/6_javascript_client.rst
```diff
@@ -229,9 +222,10 @@ It should have the configuration listed below::
     {
         ClientId = "js",
         ClientName = "JavaScript Client",
-        AllowedGrantTypes = GrantTypes.Implicit,
-        AllowAccessTokensViaBrowser = true,
-
+        AllowedGrantTypes = GrantTypes.Code,
+        RequirePkce = true,
```
veikkoeeva: I'm wondering about https://github.com/IdentityServer/IdentityServer4/pull/2856/files#diff-71f9631bfa2ca3c38575e79f74c3924fR146 and the IETF Best Practice recommendation to not to use implicit flow and move to PKCE even for web clients that if there should be a note a clear note about it? If I didn't read the diff wrong, AllowedGrantTypes = GrantTypes.Implicit was changed to AllowedGrantTypes = GrantTypes.Code.

As for context: https://medium.com/oauth-2/why-you-should-stop-using-the-oauth-implicit-grant-2436ced1c926, for further links to IETF and other explanation.

PR #2926, opened 2019-01-04
Code example updates and required NuGet package notice

ec-milan:
The following two code examples are updated:
1. Within API configuration section - 'AddJwtBearer()' method is replaced with the 'AddIdentityServerAuthentication()' method.
2. Within Creating the client section - variable 'token' is renamed to 'tokenResponse' in order to match the code example.
Also, instructions to install 'IdentityServer4.AccessTokenValidation' NuGet package, required for the 'AddIdentityServerAuthentication()' extension method, are added.

### DIFF
--- a/docs/quickstarts/1_client_credentials.rst
+++ b/docs/quickstarts/1_client_credentials.rst
@@ -154,7 +154,7 @@ Update `Startup` to look like this::
                 .AddJsonFormatters();
 
             services.AddAuthentication("Bearer")
-                .AddJwtBearer("Bearer", options =>
+                .AddIdentityServerAuthentication("Bearer", options =>
                 {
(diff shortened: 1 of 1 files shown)


leastprivilege:
It was a deliberate choice to use the JwtBearer handler.

ec-milan:
Ok, but in that case the options property 'ApiName' should be changed to 'Audience' in the given example and text explanation below should not refer to 'AddIdentityServerAuthentication()' method. I just saw that in the meantime someone else had issued a pull request for property name change, but the text is still referring to the old method.
The other code change, concerning the 'tokenResponse' variable, still applies.
I'll close this and make two separate pull request for both issues mentioned above.

PR #2928, opened 2019-01-05
Fixed variable name mismatch

ec-milan:
In given code example related to request token, variable 'token' is renamed to 'tokenResponse' in order to match other occurrences within the code example.


leastprivilege:
thanks!

PR #2932, opened 2019-01-08
Add server_error and temporarily_unavailable error codes

mpomery:
RFC 6749 section 4.1.2.1 and 4.2.2.1 define the server_error and temporarily_unavailable error codes for Authorization Code Grant.
Add them here to prevent the need for consumers to create their own constants or using incorrect error messages.

No.

**Other information**:


mpomery:
Digging some more this might not be the right thing to do.

mpomery:
Okay, I dug into this more and the problem is that we are doing our GrantValidationResult way too far from where we need it. From what I can tell, we should be responding with a HTTP 502-504.

PR #2941, opened 2019-01-11
Minor typo fix

ec-milan:
When describing implicit grant type, word 'service-side' is changed to 'server-side'.


brockallen:
thanks!

PR #2983, opened 2019-01-25
Support for ECDSA-based signatures (ES256, ES384, ES512)

larsw:
This PR adds support for both signing and verifying the following ECDSA-based JWT-algorithms; ES256, ES384 and ES512.

It also extends the discovery document and  the JWKs endpoint to reflect these changes.

The ```.AddDeveloperSigningCredential()``` has been extended with an additional (optional) parameter, ```intendedSigningAlgorithm```,
that can be used to generate ECC-based development key pairs.

```IdentityServerConstants.SigningAlgorithms.*``` can be used as arguments for this parameter.

This PR bumps the dependency on ```System.IdentityModel.Tokens.Jwt``` to ```5.4.0``` and the ```IntegrationTests``` project has been bumped to target ```net471``` instead of ```net461``` since the ```ECParameters``` class is missing in ```net46*```. This implies a breaking change for consumers of the IdentityServer which targets .NET Framework with an older version than 4.7.1.

**Other information**:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -211,3 +211,4 @@ tools/
 # IdentityServer temp files
 identityserver4_log.txt
 tempkey.rsa
+tempkey.ecdsa
--- a/host/Startup.cs
+++ b/host/Startup.cs
@@ -46,7 +46,7 @@ namespace Host
                 //.AddInMemoryClients(_config.GetSection("Clients"))
                 .AddInMemoryIdentityResources(Resources.GetIdentityResources())
                 .AddInMemoryApiResources(Resources.GetApiResources())
-                .AddDeveloperSigningCredential()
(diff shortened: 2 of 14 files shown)


leastprivilege:
Maybe it makes sense to multi-target and support that feature only if on 4.7.1 and higher...

larsw:
@leastprivilege Agree. Might be other framework features that will be needed in the future as well.

SaltyDH:
Is there anything I can do to help unblock this PR? We hit this issue today with Let's encrypt now generating ECDsa keys.

leastprivilege:
We decided to not do multi-targeting for the 2.x release. Since .NET 3.0 is around the corner, we will add EC for our 3.x 

@larsw I will keep this open as a reminder, once our 3.x branch stabilizes, it would be great if you would add the EC support there! thanks!

PR #3016, opened 2019-02-11
Adds "zero config" support for local APIs that use JWTs

leastprivilege:


leastprivilege:
We found something better.

PR #3022, opened 2019-02-11
Changed deprecated DiscoveryClient to HttpClient with GetDiscoveryDoc…

chilberto:
I noticed this unit test was providing a warning on the naming as well as the deprecated class. Just a trivial change but trying to add some value. :)


leastprivilege:
Thanks. But we don’t accept PRs to master.

PR #3024, opened 2019-02-12
Added OIDC check to Default Client Config Validator

scottbrady91:
The default implementation of `IClientConfigurationValidator` has checks that only apply to OAuth/OIDC client apps.

I've added a check on the client's `ProtocolType` to allow other protocol types to work using the default implementation.


leastprivilege:
thanks!

PR #3027, opened 2019-02-14
Add LocalAccessTokenValidation authentication scheme

Kahbazi:
https://github.com/Kahbazi/IdentityServer4.Contrib.LocalAccessTokenValidation/issues/2

To improve it we can add `AuthenticationEvents` and `RequireHttpsMetadata` to options.


leastprivilege:
The PR needs to be done against dev. thanks!

PR #3028, opened 2019-02-14
Add LocalAccessTokenValidation authentication scheme

Kahbazi:
https://github.com/Kahbazi/IdentityServer4.Contrib.LocalAccessTokenValidation/issues/2

### DIFF
--- a/src/Hosting/LocalAccessTokenValidation/LocalAccessTokenValidationDefaults.cs
+++ b/src/Hosting/LocalAccessTokenValidation/LocalAccessTokenValidationDefaults.cs
@@ -1,0 +1,17 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace IdentityServer4.Hosting.LocalAccessTokenValidation
+{
+    /// <summary>
+    /// Constants for IdentityServer authentication.
+    /// </summary>
+    public static class LocalAccessTokenValidationDefaults
+    {
+        /// <summary>
+        /// The authentication scheme
+        /// </summary>
+        public const string AuthenticationScheme = "LocalAccessTokenValidation";
+    }
+}
--- a/src/Hosting/LocalAccessTokenValidation/LocalAccessTokenValidationExtensions.cs
+++ b/src/Hosting/LocalAccessTokenValidation/LocalAccessTokenValidationExtensions.cs
@@ -1,0 +1,51 @@
+using Microsoft.AspNetCore.Authentication;
+using System;
+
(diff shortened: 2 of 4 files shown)


leastprivilege:
thanks!

Kahbazi:
You're welcome.
What more features do you think we can add?
I'm thinking Events and Required Https could be useful

danwad:
Just tried to use this but it doesn't seem to be included in the latest 2.4.0-preview1 build on NuGet.

```
<PackageReference Include="IdentityServer4" Version="2.4.0-preview1" />

...

error CS0234: The type or namespace name 'LocalAccessTokenValidation' does not exist in the namespace 'IdentityServer4.Hosting' (are you missing an assembly reference?)
```

leastprivilege:
It is 

https://github.com/IdentityServer/IdentityServer4/blob/dev/host/Startup.cs#L62

danwad:
Ah - so it is `AddLocalApiAuthentication` and not `AddLocalAccessTokenValidation`. I was trying to use it based on this PR which still calls it the latter.

Thanks

ulrichb:
@danwad ... so the idea of `AddLocalApiAuthentication()` is to enable access tokens (including reference tokens?) authentication for API Controllers which are part of the IdentityServer itself instead of using [`AddIdentityServerAuthentication`](https://github.com/IdentityServer/IdentityServer4.AccessTokenValidation/blob/2b47ee5baba65b8f72f8e4ca4365b09c2b1277c8/src/IdentityServerAuthenticationExtensions.cs#L23) which would require registering/configuring an own API client.

Is this correct?

VictorioBerra:
@ulrichb It is a thin wrapper around the [internal IS4 token validator](https://github.com/IdentityServer/IdentityServer4/blob/master/src/IdentityServer4/src/Validation/Default/TokenValidator.cs). Makes it easy to add an auth handler that invokes the ITokenValidator.

If you look at their sample controller, [it specifically targets that scheme](https://github.com/IdentityServer/IdentityServer4/blob/master/src/IdentityServer4/host/LocalApiController.cs#L13).

Keep in mind, that will new up a shit ton of services via the IoC and will access the stores and stuff too. So its really only useful if you have an API that is a part of the IdentityServer like you mentioned. Its basically like [you are calling the Introspection endpoint yourself](https://github.com/IdentityServer/IdentityServer4/blob/master/src/IdentityServer4/src/Validation/Default/IntrospectionRequestValidator.cs#L58).

Issue #2556, opened 2018-08-17
HomeController.Login does NOT handle 'cancel'


gvdonovan:
Based on the latest bits the IdentityServer4.QuickStart.UI does not handle the [Cancel] button.

brockallen:
In what way?

gvdonovan:
What should be done in mvc client to handle access denied when user cancels
login?

brockallen:
That's up to the client application (and the tech stack you use to help with the protocol). In ASP.NET Core MVC, you'd handle the error event.

gvdonovan:
I'm using the MVC app (implicit) that is provided in the samples.  Would
like to redirect to an 'Access Denied' page or something of the like if
they hit cancel.  It currently returns to http://localhost:5002/signin-oidc
with the following error:

OpenIdConnectProtocolException: Message contains error: 'access_denied',
error_description: 'error_description is null', error_uri: 'error_uri is
null'.

Is there a way to define an AccessDeniedPath in the client that Identity
Server could check?

Thanks!

brockallen:
> It currently returns to http://localhost:5002/signin-oidc with the following error: 
> OpenIdConnectProtocolException: Message contains error: 'access_denied', error_description: 'error_description is null', error_uri: 'error_uri is null'.

This is by design and up to the client to handle. Now if you're asking for the samples to be clarified, then we will gladly accept a PR for that. You'd handle it by handling the error event on OIDC options in the client.

brockallen:
Closing. If you wish to beef up the same clients, please send a PR. Thanks.

blankenbush:
You can try something like this...

```
.AddOpenIdConnect("oidc", options =>
{
    // other options here...

    options.Events = new OpenIdConnectEvents
    {
        OnRemoteFailure = (context) =>
        {
            context.Response.Redirect("/");
            context.HandleResponse();
            
            return Task.CompletedTask;
        }
    };
});

```

findmoon:
> You can try something like this...
> 
> ```
> .AddOpenIdConnect("oidc", options =>
> {
>     // other options here...
> 
>     options.Events = new OpenIdConnectEvents
>     {
>         OnRemoteFailure = (context) =>
>         {
>             context.Response.Redirect("/");
>             context.HandleResponse();
>             
>             return Task.CompletedTask;
>         }
>     };
> });
> ```

[...]

PR #3040, opened 2019-02-15
Provide a client application that handles cancel option from Identity Server/HomeController.Login

zinov:
I would like to see a running sample of how a client application handles the cancel options after a user clicks on the cancel option on the login page of the HomeController. 
Client Application: Asp.Net Core MVC

-This should not introduce a breaking change.

This was the issue I followed: 
https://github.com/IdentityServer/IdentityServer4/issues/2556

and I would like to know which is the correct way to orchestrate this from the client application perspective.

Trying something like this on the client side let me redirect to a Denied or Error page. The thing is that at this point I don't have too much information about what happened on the identity server side if the access was denied for some reason, the only thing I am getting is on the Failure property of the context the OpenIdConnectProtocolException that is a generic exception when the protocol handler encounters a protocol error. But I can get as well for an AccessDenied a generic Exception with a message: "Correlation failed". What do you suggest?

`
services .AddOpenIdConnect
{
   ...
   options.Events.OnRemoteFailure = context =>
   {
        if (context.Failure is OpenIdConnectProtocolException)
        {
            //how can I know that was caused by a denied access?
            var exception = context.Failure as OpenIdConnectProtocolException;
        }

        context.Response.Redirect("/Denied");
        context.HandleResponse();
        return Task.FromResult(0);
    };
};`


brockallen:
Sorry, we don't accept PRs on master.

As for what should happen if the user clicks 'cancel', the client should get an access_denied error passed back to the redirect_uri.

PR #3059, opened 2019-02-25
Fixed bug where the Subject was not being set on the ValidatedRequest and would not end up in the TokenIssuedSuccessEvent using Code flow

LamarLugli:
When the TokenIssuedSuccessEvent is raised on a successful code authorization, the SubjectId field is blank because the Subject property was not being set in the ValidatedRequest class.  It gets set on the other authorization flows, but the line to set it here was missing.

No

- [No existing tests exist to check anything related to this flow] Unit Tests for the changes have been added (for bug fixes / features)

**Other information**:


brockallen:
Given the merge conflicts, we added this manually. Thanks.

PR #3138, opened 2019-03-28
Make config.cs a link

thrane:
To make it easier to find the Config.cs file that is referred to I made it into a link.


brockallen:
sure :) thanks

Issue #2696, opened 2018-10-09
Document decorator pattern design with IUserClaimsPrincipalFactory


scottbrady91:
Currently, the [UserClaimsFactory](https://github.com/IdentityServer/IdentityServer4.AspNetIdentity/blob/dev/src/UserClaimsFactory.cs) class is internal. With the decorator implementation, I can pre-populate the ClaimsPrincipal, but I cannot modify it once UserClaimsFactory has finished.

brockallen:
What changes do you want to make to the claims?

scottbrady91:
I don't remember now. It was annoying at the time though. I think I hit it when migrating someone from 1.x to 2.x.

Looking over it again, maybe it would be a better idea to switching this issue/creating a new issue to add a decorator pattern example to the docs. Agreed?

ajbrun:
Hi there, it might be useful for my situation if you're able to provide a small example for your decorator implementation. I'm wanting to use AspNetIdentity's UserClaimsPrincipalFactory role implementation and I assumed adding this might take advantage of the decorator?
`services.AddTransient<IUserClaimsPrincipalFactory<ApplicationUser>, UserClaimsPrincipalFactory<ApplicationUser, ApplicationRole>>();`

However that gives me an error: "System.InvalidCastException: Object cannot be stored in an array of this type."

Looking at the code, `UserClaimsPrincipalFactory<ApplicationUser, ApplicationRole>` appears to be what I need rather than `UserClaimsPrincipalFactory<ApplicationUser>` but I haven't seen how to take advantage of it using IdentityServer4.AspNetIdentity.

brockallen:
Do we still need/wand this?

scottbrady91:
If you're offering. Otherwise, I'll PR something to http://docs.identityserver.io/en/dev/reference/aspnet_identity.html when I get a chance

brockallen:
I don't think I ever got an answer to my first question: does the existing approach of implementing the user claims principal factory and registering it with ASP.NET Identity not work?

scottbrady91:
Yes, it works.

What doesn't is when I'm migrating customers from ASP.NET Core 1.x to 2.x and having to reimplement their claims mappings. Is there a reason for `UserClaimsFactory` being internal? It wasn't in v1.

brockallen:
Oh I see why it was never public -- because we'd need to make the Decorator plumbing also public. So yes, please send a PR to the docs :)

PR #3149, opened 2019-03-31
make UserClaimsFactory public #2696

reaction1989:
#2696


brockallen:
that change causes the build to fail. not sure i want to open up the decorator classes, to be honest.

PR #3163, opened 2019-04-03
[WIP] remove hardcoded package dependency version

adamralph:

Removes the hardcoded version used for cross-solution package dependencies.

No.

**Other information**:

@leastprivilege this was the idea I mentioned. 😉

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -2,7 +2,6 @@
 
   <PropertyGroup>
     <FrameworkVersion>2.1</FrameworkVersion>
-    <IdentityServerVersion>2.4.1-*</IdentityServerVersion>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


adamralph:
Looks like the command invoking minver-cli needs to be tweaked. I guess it worked for me locally because I have minver-cli installed globally. I'll look into it and amend the commit.

adamralph:
OK, you may want to hold off on this. While it works perfectly when `build.ps1` or `build.sh` are executed, thing don't work quite so well when you build an individual solution, because `MinVerVersionOverride` is not set in that context. It *may* be possible to hack something into `Directory.Build.targets` to make that happen.

adamralph:
So I managed to hack the targets file instead, but it really is getting quite hacky now.

It now works perfectly when running either the root build scripts or one of the individual solution build scripts. But it doesn't work in Visual Studio. 😖

leastprivilege:
Also remember that I started this journey to become independent of dotnet tools in the first place? There are issue when running a tool on a different .net version than it was built against...

adamralph:
Ah yes, that does change things. OK, if I think of something else then I'll raise another PR.

Issue #3161, opened 2019-04-03
Support absolute path for DeviceVerificationUrl configuration


opolsen:
This is a suggestion related to the device flow.

I think it would be useful to be able to configure an absolute path for `DeviceVerificationUrl` in the `UserInteractionOptions` class.

The reason is that the value of the `"verification_uri"` property returned from the `/connect/deviceauthorization` endpoint, will in some cases not be a link the user can click or copy, but manually read from a monitor and written in a browser. This can be a bit tedious for the user, so it would be nice to be able to configure this to something a bit shorter and less cryptic than the Identity Server base URL and instead redirect to the Identity Server from that URL.

The current model for `UserInteractionOptions` seems to suggest that this should be possible based on this comment:
```
/// <summary>
/// Gets or sets the device verification URL.  If a local URL, the value must start with a leading slash.
/// </summary>slash.
```
With the current implementation of `DeviceAuthorizationResponseGenerator`, the `DeviceVerificationUrl` value is appended to the end of the Identity Server base URL regardless of whether it's a local or absolute path.

Do you have any thoughts on this?

scottbrady91:
The only advice I can find in the current version of the spec is:

> The URI should be short and easy to remember as end users
      will be asked to manually type it into their user-agent.

I remember there being a mention that the URL should also be trusted and recognised by the user, e.g. a domain the user recognises as owned by the IdP.

How were you thinking of shortening the URL? Would the user still be able to judge that it is legitimate?

opolsen:
Good question.

The idea was to shorten the URL to the minimum version that will still be recognized by the user.

In my case the Identity Server is hosted on sts.<domain>.com, where <domain>.com is a website the user has already visited when registering the user.

The value of `"verification_uri"` will in this case be `"https://sts.<domain>.com/device"`, where the final `/device` part is something I can configure.

Since the user might not know what `https` or `sts` is, it would be nice to be able to change this to just `<domain>.com/device` and redirect from there to the Identity Server instead, to make login as easy as possible.

scottbrady91:
Would you also use the same shortened domain for `verification_uri_complete`? Currently the IdentityServer base url and the registered device flow path are driving both values.

I think not displaying the URI scheme should be something handled by the client app, rather than in the protocol messages.

opolsen:
[...]

PR #3201, opened 2019-04-18
Allowed usage of relative and absolute verification URIs for device authorization

scottbrady91:
See #3161


brockallen:
nice, thanks

PR #3277, opened 2019-05-20
Null check for Ensure extension methods

scottbrady91:
Added null check to `EnsureLeadingSlash` and `EnsureTrailingSlash` extensions methods to be consistent with `RemoveLeadingSlash` and `RemoveTrailingSlash`.


leastprivilege:
thanks

PR #3345, opened 2019-06-18
Docs: Add API Resources appsettings.json similar to client

scotru:
Similar to the Client section, explain the method override that allows using a configuration section to load API Resources from the appsettings.json configuration file.


leastprivilege:
Thanks!

PR #3381, opened 2019-07-01
Authentication errors should redirect by default

AdamWillden:
I'm creating this PR as a discussion point ONLY 

I'm aware you're nervous of incorporating breaking changes and if agreed you'd likely roll this into your own vNext branches which I've seen discussed but do not appear to be public(?)

However, if you'd like me to create a workable request then please tell me which branch you'd like me to work off of.

Authentication errors do not appear to be handled as per the OpenId/OAuth specifications.

Firstly, please see https://openid.net/specs/openid-connect-core-1_0.html#AuthError which has a paragraph stating:

> If the End-User denies the request or the End-User authentication fails, the OP (Authorization Server) informs the RP (Client) by using the Error Response parameters defined in Section 4.1.2.1 of OAuth 2.0 [RFC6749].

Now read https://tools.ietf.org/html/rfc6749#section-4.1.2.1, specifically the following paragraph:

> If the request fails due to a missing, invalid, or mismatching
   redirection URI, or if the client identifier is missing or invalid,
   the authorization server SHOULD inform the resource owner of the
   error and MUST NOT automatically redirect the user-agent to the
   invalid redirection URI.
>
>   If the resource owner denies the access request or if the request
   fails for reasons other than a missing or invalid redirection URI,
   the authorization server informs the client by adding the following
   parameters to the query component of the redirection URI using the
   "application/x-www-form-urlencoded" format, per Appendix B:

This appears to me to state that all errors (other than for bad redirect URIs) should be returned to the client.

The OIDC spec then goes on to re-enforce this:

> Unless the Redirection URI is invalid, the Authorization Server returns the Client to the Redirection URI specified in the Authorization Request with the appropriate error and state parameters.

Again, I only made this initial PR to help portray what I understand to be required here. I'm looking for your thoughts before taking this any further as to not waste time.

Yes

**Other information**:

### DIFF
--- a/src/IdentityServer4/src/Endpoints/Results/AuthorizeResult.cs
+++ b/src/IdentityServer4/src/Endpoints/Results/AuthorizeResult.cs
@@ -78,9 +78,8 @@ namespace IdentityServer4.Endpoints.Results
                 Response.Error == OidcConstants.AuthorizeErrors.ConsentRequired ||
                 Response.Error == OidcConstants.AuthorizeErrors.InteractionRequired;
 
-            if (Response.Error != OidcConstants.AuthorizeErrors.InvalidRequestUri 
-                !isPromptNoneError ||
-                (isPromptNoneError && Response.Request?.PromptMode == OidcConstants.PromptModes.None)
+            if (Response.Error != OidcConstants.AuthorizeErrors.InvalidRequestUri &&
+                (!isPromptNoneError || (isPromptNoneError && Response.Request?.PromptMode == OidcConstants.PromptModes.None))
             )
             {
                 // this scenario we can return back to the client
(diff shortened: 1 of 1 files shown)


brockallen:
Our implementation is following this guidance: https://tools.ietf.org/html/draft-bradley-oauth-open-redirector

AdamWillden:
appreciate the link, will have a look

AdamWillden:
I have read the document over and I may have interpreted it wrong but am I correct this specification only applies where an authorisation server permits the registration of 'unauthenticated' clients.

We have a closed system where clients are registered by administrators, and clients must authenticate securely (https://tools.ietf.org/html/rfc7521). In this case I believe we do not need such a mitigation.

Is there any scope to consider an option for this behaviour? Or will I need to provide a custom endpoint handler & result?

AdamWillden:
My boss has confirmed:
>We are not an open redirector , we only redirect to predefined urls, setup out of bounds :+1:

So I would indeed be looking for an option to disable this behaviour if you'd consider such a PR (a non-breaking addition)?

AdamWillden:
Closing this non-PR in favour of this feature request: https://github.com/IdentityServer/IdentityServer4/issues/3400

PR #3382, opened 2019-07-01
add message store abstraction on authorization request params

brockallen:


VictorioBerra:
@brockallen When would someone need to use a distributed cache for the request params?

VictorioBerra:
I realize that question might be dumb. "Use a distributed cache any time you have a distributed application".

I guess my main confusion is why this would be needed when all the information is already in the query string. If my user gets bounced between servers in my farm that shouldn't matter right? So whats the point?

scottbrady91:
For the times when they are not in the query string :) [Docs](http://docs.identityserver.io/en/latest/topics/request_object.html) & [spec](https://openid.net/specs/openid-connect-core-1_0.html#JWTRequests).

VictorioBerra:
@scottbrady91 maybe im missing something then, because how would this have worked in a distributed environment before this PR?

Unless the params as a JWT in the body is somehow put back into the query string?

scottbrady91:
> how would this have worked in a distributed environment before this PR?

The ability to send authorization parameters in a JWT was added in the same release as this PR

VictorioBerra:
Ohhhhh, makes perfect sense. I missed that reading through the release notes. Thanks!

slaneyrw:
Just upgrading to v2.5 from 2.4 and noticed this change.

I was taking advantage of the fact that the handling of the CustomRedirectResult's ExecuteAsync and LoginPageResult's ExecuteAsync where identical in regards to querystring handling so I can change the login page destination *per request*  ( Return a redirect interaction response instead of login response ).

Now with the authorizationParametersMessageStore capability ( yay! we were hitting URL length issues due to state parameter ), the LoginPageResult's handling isn't replicated in the CustomRedirectPage handing and I can longer influence the destination login page.

I don't have a customization point in the AuthorizeEndpoint that would allow me to flow a custom login URL from a custom AuthorizeInteractionResponseGenerator into the login redirect - it *must* come from the IdentityServerOptions **.UserInteraction.LoginUrl** property which is a singleton and cannot be changed.

Can you please pass the InteractionResponse's RedirectUrl into the LoginPageResult so that a custom URL can be supplied in preference to the system option's value.

brockallen:
@slaneyrw open a new issue if you're asking for an enhancement, please

PR #2975, opened 2019-01-23
Incorrect secret type for missing secret in BasicAuth

sbebrys:
**What issue does this PR address?**
When in Basic Authorization client secret is missing parser set incorrect type (set SharedSecret but not NoSecret) what make exception in secret validator:
https://github.com/IdentityServer/IdentityServer4/blob/63a50d7838af25896fbf836ea4e4f37b5e179cd8/src/Validation/Default/HashedSharedSecretValidator.cs#L63

**Does this PR introduce a breaking change?**
No

**Please check if the PR fulfills these requirements**
- [ x] The commit follows our [guidelines](https://github.com/IdentityServer/IdentityServer4/blob/dev/.github/CONTRIBUTING.md)
- [ x] Unit Tests for the changes have been added (for bug fixes / features)

PR #3388, opened 2019-07-03
repro PR for Incorrect secret type for missing secret in BasicAuth #2975

brockallen:
https://github.com/IdentityServer/IdentityServer4/pull/2975

given the lack of response on the other thread


leastprivilege:
looks good

Issue #3332, opened 2019-06-11
[Device Flow] Requesting only 'offline_access' scope shows error


dudedev:
If I request only the 'offline_access' scope and access the device verification uri (https://demo.identityserver.io/device) to authorize the client I get error "Sorry, there was an error" on the page "https://demo.identityserver.io/device/UserCodeCapture".

I have used the device client given in the config [here](https://github.com/IdentityServer/IdentityServer4.Demo/blob/master/src/IdentityServer4Demo/Config.cs).

Is it a bug or I am doing something wrong?

scottbrady91:
I think all grant types would error when only asking for offline_access, since you haven't actually asked to access any protected resources.

dudedev:
I just checked Authorization Code Flow with client "server.code" (check [config](https://github.com/IdentityServer/IdentityServer4.Demo/blob/master/src/IdentityServer4Demo/Config.cs)). I requested only 'offline_access' scope and it worked. I could get access_token and refresh_token in token response.

This is also true for our Authorization Server, means it works with Authorization Code Flow but not with Device Code Flow. We did not do any special validation. So, I wanted to report this and get your feedback on this.

scottbrady91:
The below authorization request will cause IdentityServer to error. You can't generate an identity token without the openid scope, and you can't generate an access token without it being intended for a protected resource.

```
http://localhost:5000/connect/authorize
  ?client_id=mvc.hybrid.autorefresh
  &redirect_uri=http://localhost:21404/signin-oidc
  &response_type=code id_token
  &scope=offline_access
  &nonce=xyz
  &state=123
```

dudedev:
```
https://demo.identityserver.io/connect/authorize
?client_id=server.code
&scope=offline_access
&response_type=code
&redirect_uri=https://notused
```
The above authorize request works. After getting the code you can also successfully exchange the `code `for `access_token` and `refresh_token`.

So, without specifying any protected resource scope, one can get the `access_token` in this case. However, the token will be useless because it will only have `offline_access` scope.

dudedev:
So, I think we can consider this as `incorrect usage` and close this issue.

scottbrady91:
I think this *might* be a bug in IdentityServer's `ScopeValidator`.

@leastprivilege Is there a reason to allow generation of an access token with only an audience of `https://localhost:5001/resources` and scope of `offline_access`?

e.g. a payload like the following:
```
{
  "nbf": 1560688829,
[...]

PR #3391, opened 2019-07-04
Added check to scope validator for missing identity and api scopes

scottbrady91:
Fix for IdentityServer bug reported in #3332 

This stops IdentityServer issuing access tokens when only the offline_access scope is requested.


brockallen:
thanks

Issue #3329, opened 2019-06-11
[Device Flow] Token response does not contain "scope" property


dudedev:
We are using IdentityServer4 library v2.4.0 to build our Authorization Server. We implemented/enabled the device code flow. We noticed that the token response does not contain "scope" property if user has not consented to all requested scopes. However, according to the [RFC6749 section 5.1](https://tools.ietf.org/html/rfc6749#section-5.1), the "scope" property is "required" if consented scopes are different than requested scopes.

We also verified this against the demo server provided by you.

Do you plan to implement this feature soon?

brockallen:
Please have a look @scottbrady91.

scottbrady91:
@brockallen Device flow sets the authorized scopes, but I'm not seeing scope returned from the token endpoint at all: https://github.com/IdentityServer/IdentityServer4/blob/master/src/IdentityServer4/src/Endpoints/Results/TokenResult.cs#L52

scottbrady91:
So, scope is defined as a parameter on the token endpoint in the original OAuth spec. Would that ever be necessary outside of device flow?

scottbrady91:
@brockallen ping?

brockallen:
Dp we have a unit test that shows the expected result, but should be failing if the report is correct? That's the best way to go about repo'ing and then fixing.

brockallen:
merged.

PR #3392, opened 2019-07-04
Added scope param to token endpoint for device grant type

scottbrady91:
Returns scope parameter from token endpoint when the user does not consent to all scopes. Affects device flow only.

I couldn't find any tests for the token endpoint. Am I being blind?


scottbrady91:
Addresses #3329

brockallen:
Thanks for supporting the OSS!

scottbrady91:
k

Issue #3436, opened 2019-07-15
JwtRequestValidator ctor causes SimpleInjector validation error


protonyx:
We use SimpleInjector in tandem with the built-in Microsoft DI (https://simpleinjector.readthedocs.io/en/latest/aspnetintegration.html) to provide more advanced (and opinionated) DI scenarios. During our startup flow, we call a `Verify `method on the SimpleInjector container which walks through the dependency tree and ensures that all dependencies are constructable.

When SimpleInjector gets to JwtRequestValidator, an exception is thrown because `contextAccessor.HttpContext` is null, which is expected behavior when you are calling this method outside the scope of a request. 

This issue is not seen when using the default Microsoft DI container, because it does not try to construct all classes known to the container. This capability of SimpleInjector allows us to fail fast and detect DI configuration-related issues on startup. 

This is not a high priority issue, since it has limited impact to an optional feature of a non-standard DI container.

leastprivilege:
Is SimpleInjector clever enough to deal with ArgumentNullExceptions?

protonyx:
I did some testing, and it looks like it is not. SimpleInjector assumes that all registered classes are constructable. `HttpContextAccessor` is constructable without providing an `HttpContext` (https://github.com/aspnet/HttpAbstractions/blob/master/src/Microsoft.AspNetCore.Http/HttpContextAccessor.cs), so a call to `HttpContextAccessor.HttpContext` without setting the context will return null.

brockallen:
So where are we on this issue?

armitagemderivitec:
I'm using Autofac in tandem with Microsoft DI and I'm getting the same error in my integration tests for controller instantiation. Each of my controllers that inject IIdentityServerInteractionService cannot now be tested. Any ideas?

ChristianKuhre:
I am experiencing the same issue when using Simple Injector and updating to 2.5.0, so it is kinda keeping us from updating from 2.4.0. Maybe the HttpContext access could be moved away from the constructor?

leastprivilege:
Will be fixed here

https://github.com/IdentityServer/IdentityServer4/pull/3517

protonyx:
Thank you, that change fixes our issue.

PR #3437, opened 2019-07-15
check that HttpContext is set before setting AudienceUri (#3436)

protonyx:
Allows the `JwtRequestValidator` to be constructable when the HttpContext is not set because it is constructed outside the context of a request. This was discovered by using SimpleInjector `Verify` that attempts to construct all dependencies to weed out DI configuration-related errors.

Addresses issue #3436 

No


leastprivilege:
I am not quite sure I agree with this approach. What's actually missing in our code is a check for null and throw an exception if HttpContext is null...

Is this the very first time you run into such a situation with the simple injector "validation" feature?

leastprivilege:
Will be fixed here

https://github.com/IdentityServer/IdentityServer4/pull/3517

for 2.5.2

PR #3463, opened 2019-07-23
Easier Authorization Code extensibility

leastprivilege:
e.g. to transport additional values from the authorization request to the token request.

### DIFF
--- a/src/Storage/src/Models/AuthorizationCode.cs
+++ b/src/Storage/src/Models/AuthorizationCode.cs
@@ -108,5 +108,13 @@ namespace IdentityServer4.Models
         /// The code challenge method
         /// </value>
         public string CodeChallengeMethod { get; set; }
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


src/Storage/src/Models/AuthorizationCode.cs
```diff
@@ -108,5 +108,13 @@ public class AuthorizationCode
...
         public string CodeChallengeMethod { get; set; }
+
+        /// <summary>
+        /// Gets or sets properties
+        /// </summary>
+        /// <value>
+        /// The properties
+        /// </value>
+        public IDictionary<string, object> Properties { get; set; } = new Dictionary<string, object>();
     }
```
brockallen: Why `object` in the `Dictionary<string, object>`? I worry about the values not round tripping.

leastprivilege: done

PR #3464, opened 2019-07-23
Easier support for impersonating clients

leastprivilege:
useful for token exchange scenarios


leastprivilege:
made the changes we discussed

PR #3488, opened 2019-07-27
Update Startup.cs

epoyraz:
This PR moves just the call from UseStaticFiles before UseRouting in the Startup

According the migrating guide (source: https://docs.microsoft.com/en-us/aspnet/core/migration/22-to-30?view=aspnetcore-2.2&tabs=visual-studio) UseStaticFiles should be called before UseRouting

> If the app calls UseStaticFiles, place UseStaticFiles before UseRouting

no

**Other information**:

### DIFF
--- a/src/IdentityServer4/host/Startup.cs
+++ b/src/IdentityServer4/host/Startup.cs
@@ -88,13 +88,13 @@ namespace Host
 
         public void Configure(IApplicationBuilder app)
         {
+            app.UseStaticFiles();
             app.UseRouting();
             app.UseMiddleware<Logging.RequestLoggerMiddleware>();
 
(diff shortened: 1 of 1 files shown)


leastprivilege:
Thanks - I did it slightly different.

Did you find anywhere an explanation why the routing MW should come after static files? for performance reasons?

My understanding was that you put the routing MW at the point in the pipeline from where you want routing information to be available, and since our request logger also logs the route name, I put it very early.

leastprivilege:
Maybe @filipw knows?

epoyraz:
I don't know the explaination.
I tried to google it, but i didn't find any clues.

PR #3491, opened 2019-07-29
fix JS for automatic signout redirect

### LINKED ISSUE #3475: Quickstart.UI signout-redirect.js does not need load event
Recent change in `signout-redirect.js` file in Quickstart.UI replaces `window.addEventListener()` with `$()` [1]. I think none is necessary actually.

As far as I see, in all scenarios the script is included after the relevant anchor element in the DOM. Therefore it's safe to just query for the element. In fact this is pretty much exactly what is done in the sibling file `signin-redirect.js` [2].

[1] https://github.com/IdentityServer/IdentityServer4.Quickstart.UI/commit/bedfed33462f4223bee4a0519c361c59aa8d3753
[2] https://github.com/IdentityServer/IdentityServer4.Quickstart.UI/blob/bedfed33462f4223bee4a0519c361c59aa8d3753/wwwroot/js/signin-redirect.js

brockallen:
fixes #3475

### DIFF
--- a/src/IdentityServer4/host/wwwroot/js/signout-redirect.js
+++ b/src/IdentityServer4/host/wwwroot/js/signout-redirect.js
@@ -1,4 +1,4 @@
-﻿﻿$(function () {
+﻿$("iframe.signout").load(function () {
     var a = document.querySelector("a.PostLogoutRedirectUri");
     if (a) {
         window.location = a.href;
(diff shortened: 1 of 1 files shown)


brockallen:
This puts it back to how we had it prior to 2.5.0.

PR #3513, opened 2019-08-03
Set typ header for access tokens

leastprivilege:
...this would also allow removing the `../resources` audience.

TODO: 

* [x] check typ in the built-in token validator.
* [x] make token type value configurable on `IdentityServerOptions`
* [x] remove the old resources audience?


reaction1989:
@leastprivilege The AccessTokenValidation with IdentityServer3.AccessTokenValidation will only throw errors.

I ran into that issue. Main Problem is the removal of /ressources because the lib uses that for the validation of audience

leastprivilege:
Both the custom typ and existence of /resource can be configured. It's all documented.

Issue #3535, opened 2019-08-12
Uris with custom schemes are not working in 2.5.2.


pavinan:
This update broke custom scheme urls. (Ex: "capacitor://localhost", "ionic://localhost"). Ionic app in ios or android renders pages with those urls inside webview.
According to [oauth rfc 6749 section 3.1.2](https://tools.ietf.org/html/rfc6749#section-3.1.2) custom schemes are allowed.

### Issue / Steps to reproduce the problem

Add some CORS origins with custom schemes. (Ex: "capacitor://localhost", "ionic://localhost")

### Relevant parts of the log file

```
Invalid client configuration for client sampleclient: AllowedCorsOrigins contains invalid origin: ionic://localhost
```

brockallen:
>  This update broke custom scheme urls. (Ex: "capacitor://localhost", "ionic://localhost"). Ionic app in ios or android shows page in those urls.

Why are you putting those in `AllowsCorsOrigins`? They don't need to be in there.

pavinan:
No, ionic creates some kind of server inside mobile and opens it in webview. That URL happens to be "capacitor://localhost".

Docs: https://ionicframework.com/docs/faq/cors

adozenlines:
Still an issue in version 3.0.0 with custom schemes.

adozenlines:
Mobile Applications that perform OAuth have the ability to register custom URL schemes. When such a scheme is specified in the call back URL of the request, the identity server doesn't redirect to the custom scheme and as a result, the mobile application never receives the access token.

pavinan:
@adozenlines what framework are you using? Is it Ionic?

pavinan:
Hi @brockallen this issue is blocking us from updating to .Net Core 3.0. Is there any alternative and can you check PR #3559. Thank you.

pavinan:
@brockallen I have overridden ValidateAllowedCorsOriginsAsync in DefaultClientConfigurationValidator. Everything is working good.

brockallen:
Tracked and fixed here:  #3880

PR #3559, opened 2019-08-25
Removed scheme checking in cors origin validation.

pavinan:
Allows custom schemes in CORS Uri.

No

**Other information**:
Related to #3535


MZap:
I was curious if this also needs to get fixed:
[EndSessionCallbackResult](https://github.com/IdentityServer/IdentityServer4/blob/97317677d714b31d7ae8b20b7de0810e81df74c5/src/IdentityServer4/src/Endpoints/Results/EndSessionCallbackResult.cs#L65)

I noticed it makes a call to GetOrigin [GetOrigin](https://github.com/IdentityServer/IdentityServer4/blob/97317677d714b31d7ae8b20b7de0810e81df74c5/src/IdentityServer4/src/Extensions/StringsExtensions.cs#L246)

` private void AddCspHeaders(HttpContext context)
        {
            if (_options.Authentication.RequireCspFrameSrcForSignout)
            {
                string frameSources = null;
                var origins = _result.FrontChannelLogoutUrls?.Select(x => x.GetOrigin());
                if (origins != null && origins.Any())
                {
                    frameSources = origins.Distinct().Aggregate((x, y) => $"{x} {y}");
                }

                // the hash matches the embedded style element being used below
                context.Response.AddStyleCspHeaders(_options.Csp, "sha256-u+OupXgfekP+x/f6rMdoEAspPCYUtca912isERnoEjY=", frameSources);
            }
        }`

It looks like if a null is returned from GetOrigin() this might impact somethings downstream



It would be nice to get a status update on this PR as well. We are about to move to use ID4 and don't want to fork this repo to avoid customization

pavinan:
If removing scheme validation is going to be a security concern. It is better to give an option like a boolean property ``SuppressSchemeValidation``.

brockallen:
> It would be nice to get a status update on this PR as well. We are about to move to use ID4 and don't want to fork this repo to avoid customization

Rarely do you need to fork this repo, as there are numerous extensibility points.

MZap:
Hi Brock, this is very true. We are taking advantage of the DI framework and replacing the classes we need. Thanks!

PR #3490, opened 2019-07-28
ensure SDK version for MinVer is installed

adamralph:
🤞

PR #3630, opened 2019-09-14
move entirely to netcoreapp3.0

adamralph:

In `master`, an extra step [was introduced](https://github.com/IdentityServer/IdentityServer4/pull/3490) to install the .NET Core 2.x SDK during the CI build. One of the reasons was that MinVer 1.x requires a .NET Core 2.x SDK to be present on the machine.

Starting with MinVer 2.0.0 (currently alpha 2), a .NET Core 2.x SDK is no longer required.

However, a number of projects in the solution (e.g. many samples) are still using `netcoreapp2.x` so it wasn't just MinVer that required a .NET Core 2.x SDK to be present. I've upgraded all remaining projects from `netcoreapp2.x` to `netcoreapp3.0`, but I'm not sure if that has any unintended consequences.

I'm not entirely sure. The upgrade of all remaining projects from `netcoreapp2.x` to `netcoreapp3.0` would need to be examined.

- ~Unit Tests for the changes have been added (for bug fixes / features)~ - _n/a_

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -8,11 +8,6 @@ jobs:
     vmImage: 'ubuntu-16.04'
   steps:
   - task: UseDotNet@2
-    displayName: 'Install .NET Core 2.x SDK'
-    inputs:
-      packageType: 'sdk'
-      version: '2.x'
-  - task: UseDotNet@2
     displayName: 'Install .NET Core 3.x SDK'
     inputs:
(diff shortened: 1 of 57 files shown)


adamralph:
I've just realised that the samples aren't used for CI, so I'm withdrawing this in favour of https://github.com/IdentityServer/IdentityServer4/pull/3631.

PR #3657, opened 2019-09-23
Check user active/passive from IPassiveable first on ProfileService

mehyaa:
Check if the user is active from IPassiveable interface if entity implements it.

AspNetIdentity User entity has no active/passive property but on some scenario you need capability of making user active/passive.


brockallen:
I think this adds some bloat. I'd rather have you just derive and override the IsActive for your custom scenarios.

Issue #2067, opened 2018-02-12
Why does the IdentityServer4.EntityFramework project not use asynchronous EF methods?


neyromant:
Hi, I'm surprised that all calls to Db Context are synchronous. And I'm ready to replace them with asynchronous ones.
I wanted to know why it was not done before?

Is there any good reason for this?

neyromant:
UPD:
@brockallen I found your use_async_ef branch (https://github.com/IdentityServer/IdentityServer4.EntityFramework/compare/use_async_ef)

Is there any reasons why it does not merge into the main branch?

brockallen:
We've had issues with the underlying EF implementation under load.

neyromant:
Currently we can see the following behavior - under load and with a long ping to DB, all application's threads get busy. Kestrel successfully processes connections, but the application is not responsible for sufficient time.

neyromant:
Can we find out what kind of problems you had? Maybe you have stress tests?

neyromant:
I have another question - why do not you use AsNoTracking for read-only queries? Is there any reason for this?

neyromant:
@brockallen I added AsNoTracking for readonly queries - https://github.com/IdentityServer/IdentityServer4.EntityFramework/pull/136

brockallen:
What does `AsNoTracking` do? How does it affect different providers?

TomCJones:
Is this designed to hold the value of DNT - or DoNotTrack from the HTTP header?

DNT would mean (inter alia) that any existing OP signin cannot store or release information about the sites visited. (Obviously the OP does record all the clients that have been registered for a user, at least if pairwise sub is implemented.) It should imply that no client or external provider can get any information about signin to other sites. Vanilla OpenID Connect should provide this by default. API grants might leak this information if not careful.

neyromant:
@brockallen  AsNoTracking avoids the overhead of setting up change tracking for each entity instance
> What does AsNoTracking do? How does it affect different providers?

brockallen:
Since this store API is designed as read-only, would it make more sense to set this globally? https://docs.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.changetracking.changetracker.querytrackingbehavior?view=efcore-2.0#Microsoft_EntityFrameworkCore_ChangeTracking_ChangeTracker_QueryTrackingBehavior?

scottbrady91:
[...]

PR #3674, opened 2019-09-25
EF Async

AliBazzi:

This PR tries to address the synchronous calls over EF, and instead use the asynchronous calls 
#2067 

No


src/EntityFramework.Storage/src/Stores/DeviceFlowStore.cs
```diff
@@ -60,43 +59,41 @@ public class DeviceFlowStore : IDeviceFlowStore
...
         /// <returns></returns>
-        public virtual Task StoreDeviceAuthorizationAsync(string deviceCode, string userCode, DeviceCode data)
+        public virtual async Task StoreDeviceAuthorizationAsync(string deviceCode, string userCode, DeviceCode data)
         {
             Context.DeviceFlowCodes.Add(ToEntity(data, deviceCode, userCode));
 
-            Context.SaveChanges();
-
-            return Task.FromResult(0);
+            await Context.SaveChangesAsync();
```
khellang: Coulnd't this be returned directly, without compiling a state machine?

AliBazzi: you mean something like:
`return context.SaveChangeAsync()` ?

khellang: Yes and remove the `async` keyword 😄

AliBazzi: good suggestion, I didn't know that would have performance impact
I will look again on the places where this could be implemented

khellang: Here's an example of a compilation with and without `async`/`await`; https://gist.github.com/khellang/6e7b3bd197657112714c.

jepperc: Doing this hides where the exception happens.. don't remove async / await unless it's extremely performance critical. And since this has a DB call, it isn't really.

AliBazzi: @jepperc point made sense to me, reverted the changes

brockallen:
Closing since this PR is out of date and I added another. Sorry for not remembering about it first.

PR #3692, opened 2019-10-01
Removed obsolete code

khellang:
I stumbled over these while looking into endpoint routing.

I guess the ship has sailed for 3.0, but this PR might serve as a reminder for the next major update?


leastprivilege:
thanks!

Yeah - we were a bit rushed with the 3.0 update... ;)

brockallen:
Yay, my comments back then were useful. Thanks!

PR #3713, opened 2019-10-09
Add more comments to 'claims' parameter of GrantValidationResult constructor.

voroninp:
I still believe `claims` parameter of `GrantValidationResult` is somewhat confusing. It was yet another gotcha for me. When access token is being refreshed no `GrantValidationResult` is involved in the process, and it is the responsibility of `IProfileSevice` to retrieve all required claims.

No

- [ +] The commit follows our [guidelines](https://github.com/IdentityServer/IdentityServer4/blob/master/.github/CONTRIBUTING.md)
- [ Not Relevant] Unit Tests for the changes have been added (for bug fixes / features)

**Other information**:
This might seem to be a questionable expansion of the comments, so decide yourself whether it is worth adding it.

### DIFF
--- a/src/IdentityServer4/src/Validation/Models/GrantValidationResult.cs
+++ b/src/IdentityServer4/src/Validation/Models/GrantValidationResult.cs
@@ -77,7 +77,8 @@ namespace IdentityServer4.Validation
         /// <param name="subject">The subject claim used to uniquely identifier the user.</param>
         /// <param name="authenticationMethod">The authentication method which describes the custom grant type.</param>
         /// <param name="claims">Additional claims that will be maintained in the principal. 
-        /// If you want these claims to appear in token, you need to add them explicitly in your custom implementation of <see cref="Services.IProfileService"/> service.</param>
+        /// If you want these claims to appear in token, you need to add them explicitly in your custom implementation of <see cref="Services.IProfileService"/> service.
+        /// Beware as well that these claims won't present in refreshed access token because <see cref="GrantValidationResult"/> is not used in refresh_token flow.</param>
         /// <param name="identityProvider">The identity provider.</param>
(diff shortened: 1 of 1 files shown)


leastprivilege:
I don't follow your comment about refresh tokens - a refreshed token is an exact copy of the original token (unless you run the profile service again).

voroninp:
When I have a custom flow, I can add claims to `GrantValidationResult`.

For these claims to appear in access token, `ProfileService` must copy them to issued token.

However, when I refresh access token the code of the custom flow is not called at all, hence, these “additional” claims won’t appear  in access token, because they simply won’t present in subject claims.

leastprivilege:
As I said - a refresh token is a copy of the original access token. Unless you specifically configure it to run the profile service on refresh requests - custom claims are included automatically.

voroninp:
How come? If access tokens are not persisted?

leastprivilege:
But refresh tokens (and they associated content) are. I just added a test, just to prove I am right (because I couldn't remember how it worked ;)

voroninp:
Hm, it's weird, because for us user profile service is getting called on token refresh and there are no custom claims there.

leastprivilege:
https://github.com/IdentityServer/IdentityServer4/commit/d2470e01dc79161fd61ffd2d7b358d7ede1280bc

leastprivilege:
Maybe I am testing something different, than what you are talking about ... have a look and let me know.

voroninp:
Ok, but maybe how we do it should not work at all.
Legacy implementation returns absolutely all claims of the authenticated user. Custom claims are persisted in `UserClaims` table.

So for the requested scopes `openid offline_access` we return all claims.

leastprivilege:
OK - seems there is no issue with IdentityServer. I will close this.

Issue #2569, opened 2018-08-23
Consider using new endpoint routing features in ASP.NET Core 3.0


brockallen:

JoeyHerrington:
Meanwhile, do you happen to have a list of settings on-hand that we can use?
```csharp
services.AddIdentityServer(options => {
    options.UserInteraction.LoginUrl = "/Identity/Account/Login";
    // etc...
})
```
right now, it's just trial and error with more of the latter!

brockallen:
You mean like docs? 

http://docs.identityserver.io/en/release/topics/startup.html
docs.identityserver.io/en/release/reference/options.html

JoeyHerrington:
Sorry, I found exactly what I was looking for right after posting my comment and forgot to delete my question!  Thanks!

Blizzard1978:
I used new "AddControllersWithViews();"  and it crashed in apsnetcore 3.0 pr9 on the autorization attribute in consent controller fixed it by adding "app.UseAutorization();" after the "app.UseIdentityServer();" in the pipeline, this is because use authentication and use autorization has been pulled appart.

the--dud:
Thanks for the quick reply. This is related to our previous issue https://github.com/IdentityServer/IdentityServer4/issues/4047 , which relates to our efforts to support encrypted responses (encrypted ID token + encrypted userinfo response).

brockallen:
IIRC there is a way to override our existing endpoints -- it's just a matter of being configured in DI in the right order.

PR #3715, opened 2019-10-09
Add bare minimum to run on endpoint routing

khellang:
This will allow users to use either endpoint routing (by using `MapIdentityServer`) or regular middleware (by calling `UseIdentityServer`).

I've tried to keep this free from breaking changes, but it's still missing some stuff. I guess we should determine how far we're willing to go with breaking changes and what the migration story should be.

### DIFF
--- a/src/IdentityServer4/host/Startup.cs
+++ b/src/IdentityServer4/host/Startup.cs
@@ -96,12 +96,13 @@ namespace Host
 
             app.UseRouting();
             app.UseMiddleware<Logging.RequestLoggerMiddleware>();
-            app.UseIdentityServer();
 
+            app.UseCors();
             app.UseAuthorization();
 
             app.UseEndpoints(endpoints =>
             {
+                endpoints.MapIdentityServer();
                 endpoints.MapDefaultControllerRoute();
             });
         }
--- a/src/IdentityServer4/src/Configuration/IdentityServerApplicationBuilderExtensions.cs
+++ b/src/IdentityServer4/src/Configuration/IdentityServerApplicationBuilderExtensions.cs
@@ -27,7 +27,7 @@ namespace Microsoft.AspNetCore.Builder
         /// <returns></returns>
         public static IApplicationBuilder UseIdentityServer(this IApplicationBuilder app, IdentityServerMiddlewareOptions options = null)
         {
-            app.Validate();
+            app.ApplicationServices.Validate();
 
             app.UseMiddleware<BaseUrlMiddleware>();
 
@@ -47,15 +47,15 @@ namespace Microsoft.AspNetCore.Builder
             return app;
         }
 
-        internal static void Validate(this IApplicationBuilder app)
+        internal static void Validate(this IServiceProvider provider)
         {
-            var loggerFactory = app.ApplicationServices.GetService(typeof(ILoggerFactory)) as ILoggerFactory;
+            var loggerFactory = provider.GetService(typeof(ILoggerFactory)) as ILoggerFactory;
             if (loggerFactory == null) throw new ArgumentNullException(nameof(loggerFactory));
 
             var logger = loggerFactory.CreateLogger("IdentityServer4.Startup");
             logger.LogInformation("Starting IdentityServer4 version {version}", typeof(IdentityServerApplicationBuilderExtensions).Assembly.GetName().Version.ToString());
 
-            var scopeFactory = app.ApplicationServices.GetService<IServiceScopeFactory>();
+            var scopeFactory = provider.GetService<IServiceScopeFactory>();
 
             using (var scope = scopeFactory.CreateScope())
             {
--- a/src/IdentityServer4/src/Configuration/IdentityServerEndpointDataSource.cs
+++ b/src/IdentityServer4/src/Configuration/IdentityServerEndpointDataSource.cs
@@ -1,0 +1,86 @@
+using System.Collections.Generic;
+using System.Threading;
+using IdentityServer4.Extensions;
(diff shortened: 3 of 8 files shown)


src/IdentityServer4/host/Startup.cs
```diff
@@ -96,12 +96,13 @@ public void Configure(IApplicationBuilder app)
 
             app.UseRouting();
             app.UseMiddleware<Logging.RequestLoggerMiddleware>();
-            app.UseIdentityServer();
 
+            app.UseCors();
```
khellang: If you don't add the CORS middleware and try to invoke an endpoint that requires CORS, you'll get this big fat error:

image

src/IdentityServer4/src/Configuration/IdentityServerApplicationBuilderExtensions.cs
```diff
@@ -47,15 +47,15 @@ public static IApplicationBuilder UseIdentityServer(this IApplicationBuilder app
             return app;
         }
 
-        internal static void Validate(this IApplicationBuilder app)
+        internal static void Validate(this IServiceProvider provider)
```
khellang: Moved this onto `IServiceProvider` so it could be shared with `MapIdentityServer`.

src/IdentityServer4/src/Configuration/IdentityServerEndpointDataSource.cs
```diff
@@ -0,0 +1,86 @@
...
+            _options = options;
+            _requestDelegate = CreateRequestDelegate(options);
+            _endpoints = ConvertEndpoints(endpoints);
+        }
+
+        public override IReadOnlyList<Endpoint> Endpoints => _endpoints;
+
+        public override IChangeToken GetChangeToken() => new CancellationChangeToken(CancellationToken.None);
+
+        private IReadOnlyList<Endpoint> ConvertEndpoints(IEnumerable<IdentityServer4.Hosting.Endpoint> endpoints)
```
khellang: This is where the magic happens. This method loops through the existing IdentityServer endpoints and "converts" them to ASP.NET Core endpoints. All endpoints created here use the same `RequestDelegate` (created in `CreateRequestDelegate` below).

src/IdentityServer4/src/Configuration/IdentityServerEndpointDataSource.cs
```diff
@@ -0,0 +1,86 @@
...
+                // TODO: Add HTTP method metadata.
+                // builder.Metadata.Add(new HttpMethodMetadata(endpoint.Methods));
+
+                result.Add(builder.Build());
+            }
+
+            return result;
+        }
+
+        private static RequestDelegate CreateRequestDelegate(IdentityServerOptions options)
```
khellang: This `RequestDelegate` should more or less reflect the middlewares added here:

https://github.com/IdentityServer/IdentityServer4/blob/660f515469366861423c6d7b88243448442030cf/src/IdentityServer4/src/Configuration/IdentityServerApplicationBuilderExtensions.cs#L28-L48

It's still missing the auth middleware and `MutualTlsTokenEndpointMiddleware`. It shouldn't need the CORS middleware as the CORS policy should be enforced by adding the `EnableCorsAttribute` to the endpoint above.

I just pulled out the behavior of the `BaseUrlMiddleware` and `IdentityServerMiddleware` to some static methods so it could be used in both places at the same time. In the future, both middlewares will be gone and the code could be moved in here (or elsewhere).

src/IdentityServer4/src/Configuration/IdentityServerEndpointDataSource.cs
```diff
@@ -0,0 +1,86 @@
...
+
+                var builder = new RouteEndpointBuilder(_requestDelegate, pattern, order: 0) { DisplayName = endpoint.Name };
+
+                builder.Metadata.Add(new RouteNameMetadata(endpoint.Name));
+                builder.Metadata.Add(new EndpointNameMetadata(endpoint.Name));
+                builder.Metadata.Add(new IdentityServerEndpointMetadata(endpoint));
+                builder.Metadata.Add(new EnableCorsAttribute(_options.Cors.CorsPolicyName));
+
+                // TODO: Add HTTP method metadata.
+                // builder.Metadata.Add(new HttpMethodMetadata(endpoint.Methods));
```
khellang: Once the endpoints have the correct HTTP method metadata and the middleware is gone, the router will take care of returning a 405 response for the wrong methods. That means checks like these can be removed:

https://github.com/IdentityServer/IdentityServer4/blob/660f515469366861423c6d7b88243448442030cf/src/IdentityServer4/src/Endpoints/DiscoveryEndpoint.cs#L38-L43

leastprivilege:
Thanks for doing this! We will have a look ASAP!

brockallen:
moving to future. thx.

khellang:
Are you dropping plans for endpoint routing? 🤔

leastprivilege:
Wird - I cannot remember closing it...

leastprivilege:
Ah...I deleted the master branch. lol

khellang:
> Ah...I deleted the master branch. lol

😂 Can recommend this if you need to do more repos; https://github.com/mheap/github-default-branch

leastprivilege:
thanks! ;)

Do you want to re-create the PR so it is not lost?

khellang:
> Do you want to re-create the PR so it is not lost?

I don't have time to finish it right now, so it might as well just stay closed. There's already #2569 to track the work 😄

leastprivilege:
no problem.

PR #3724, opened 2019-10-12
small text change in grant types

damienbod:


leastprivilege:
The code already has a comment wrt confidential or not. 

If you feel there should a separate paragraph or a not about when to use client credentials - feel free to suggest some text.

damienbod:
true, just though it might help to make this clear in the text as well, but you also mention this early in the doccs, so

PR #3809, opened 2019-11-11
Updated EF Quickstart

scottbrady91:
- Made EF quickstart the 4th quickstart
- Continues from previous quickstart and adds EF to IdentityServer
- Updated quickstart intro & wording to be more inline with new quickstarts


leastprivilege:
thanks!

Issue #3839, opened 2019-11-19
Issue a warning or throw an exception for not completely configured ResourceOwnePassword flow.


voroninp:
**Is your feature request related to a problem? Please describe.**
It's easy to forget to register custom implementation of `IResourceOwnerPasswordValidator` when client is allowed to use resource owner credentials flow.

**Describe the solution you'd like**
I would like to see at least a warning, which says that clients A,B and D can use the flow, but it is not completely configured. Throwing an exception is definitely a breaking change, but failing at startup is much better than just log message `flow is not supported`.

voroninp:
Though it is weird because commit e504a423 contains line

```CSharp
_logger.LogInformation("Resource owner password credential type not supported. Configure an IResourceOwnerPasswordValidator.");
```

However I never saw the part about `IResourceOwnerPasswordValidator`

voroninp:
Ah, I got it. We log only errors, so this is not shown in log, but failed grant validation is logged as error, however, it contains string which is coming from `ConvertTokenErrorEnumToString(error);` So information is partially lost. Error description is returned as part of the response, yet it is not logged.

Ugly solution is to add another enum value `TokenRequestErrors.ResourceOwnerPasswordValidatorNotImplemented` and create correct string from it.

leastprivilege:
As I said before - extending the client config validator is the right way to do that.

PR #3840, opened 2019-11-19
Add remarks about custom implementation of IResourceOwnerPasswordValidator for the corresponding flow.

voroninp:
https://github.com/IdentityServer/IdentityServer4/issues/3839

No

- [ Yes] The commit follows our [guidelines](https://github.com/IdentityServer/IdentityServer4/blob/master/.github/CONTRIBUTING.md)
- [ N/A] Unit Tests for the changes have been added (for bug fixes / features)

### DIFF
--- a/src/IdentityServer4/src/Models/GrantTypes.cs
+++ b/src/IdentityServer4/src/Models/GrantTypes.cs
@@ -31,9 +31,15 @@ namespace IdentityServer4.Models
         public static ICollection<string> ClientCredentials =>
             new[] { GrantType.ClientCredentials };
 
+        /// <summary>
+        /// As Identity Server does not know how to validate user password you need to implement and register the <see cref="IdentityServer4.Validation.IResourceOwnerPasswordValidator"/> interface.
+        /// </summary>
         public static ICollection<string> ResourceOwnerPassword =>
             new[] { GrantType.ResourceOwnerPassword };
 
+        /// <summary>
+        /// As Identity Server does not know how to validate user password you need to implement and register the <see cref="IdentityServer4.Validation.IResourceOwnerPasswordValidator"/> interface.
+        /// </summary>
         public static ICollection<string> ResourceOwnerPasswordAndClientCredentials =>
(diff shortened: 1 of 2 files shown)


leastprivilege:
is 

https://identityserver4.readthedocs.io/en/latest/topics/resource_owner.html 

not good enough?

voroninp:
@leastprivilege  Yep, but it is literally RTFM =)  
Why not let the code tell the story and even better signal about potential problem?

While migrating from IS3 to IS4 we simply forgot to include the call to `AddResourceOwnerValidator` (btw why not ResourceOwnerPasswordValidator`?) =(

leastprivilege:
Because we always have a resource owner password validator registered (to keep DI happy) - just by default one that does nothing - so we cannot easily detect at startup if this is a valid config or not.

voroninp:
@leastprivilege How come? You know the exact type of the default validator, and in case the resolved instance is of this type, IS could warn:
> Hey, man, I see you didn't override default implementation of `IRsourceOwnerPasswordValidator`, but your client `Client` supports `resource owner password` flow. That's not ok.

leastprivilege:
You wanted this at startup time - we are not scanning client configuration at startup time.

voroninp:
Then at least some XML Doc would be helpful, I assume.

voroninp:
BTW, Simple Injector DI container has method `Verify`, maybe it is worth having such diagnostics feature for IS as well.

leastprivilege:
We cannot scan client config - check the `IClientStore`.

There is a configuration validator that could be used for your purpose. Feel free add that logic via a PR.

leastprivilege:
https://github.com/IdentityServer/IdentityServer4/blob/master/src/IdentityServer4/src/Validation/Default/DefaultClientConfigurationValidator.cs

voroninp:
@leastprivilege Yep, my bad, this will work fine with InMemory clients only.

voroninp:
@leastprivilege  But what about XML Doc?

leastprivilege:
Closing this in favour of the config validator

PR #3870, opened 2019-12-02
Replaces my name with my company name

henriksen:
Since it's my company that pays for the sponsorship, it's only fair that their name should be on the sponsorship list.

Corrects a name in the sponsorship list

No, text change only


leastprivilege:
done, thanks

PR #3875, opened 2019-12-04
make TokenCleanupInterval more realistic

LindaLawton:
TokenCleanupInterval of 30 seconds seams a bit extreme to me.  Wouldn't it be better to show the default of 3600 and let them decide if they want to set it down from there?

None being lazy with documentation

none 

**Other information**:

### DIFF
--- a/docs/reference/ef.rst
+++ b/docs/reference/ef.rst
@@ -80,7 +80,7 @@ To use the operational store support, use the ``AddOperationalStore`` extension 
 
                 // this enables automatic token cleanup. this is optional.
                 options.EnableTokenCleanup = true;
-                options.TokenCleanupInterval = 30; // interval in seconds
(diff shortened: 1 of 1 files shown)


brockallen:
The default is 3600. The intent was to show how it could be set. I assumed the end developer would have the sense to set it to what made sense.

LindaLawton:
The issue with you assumption is that you assume that they know what it does in the first place and aren't just copying it.

I just started at a company and the first thing they asked me is if it should be running so often.

I normally only run it once a day.   3600 does show it could be changed.

As always its up to you but I think it should be set to a more reasonable number for the average company.  30 seams extreme

brockallen:
I suspect that line was copies from my testing code, which made sense to run it that often.

Issue #3533, opened 2019-08-10
Consider breaking change for supporting multiple signing keys


scottbrady91:
Currently, only a single signing key can be registered, and therefore only a single signing algorithm supported.

To improve compatibility, we could allow multiple signing keys & algorithms to be registered. A default algorithm could be set and then overridden on a per client basis.

scottbrady91:
(one for v4)

brockallen:
Merged.

PR #3892, opened 2019-12-10
V4: Multiple signing keys

leastprivilege:
https://github.com/IdentityServer/IdentityServer4/issues/3533


leastprivilege:
@brockallen need help with the EF mappings. Maybe we store the signing algorithm list as a space separated string instead of introducing an extra table...

Issue #3552, opened 2019-08-20
No default instance for IdentityServer4.Stores.IAuthorizationParametersMessageStore


bzamora:
IdentityServer 2.5.2

When using StructureMap as DI (v1.4.0), receive an error for missing default instance of IdentityServer4.Stores.IAuthorizationParametersMessageStore

Can work around by creating a dummy implementation or downgrading to IdentityServer 2.4.0

### Issue / Steps to reproduce the problem

Using StructureMap as DI

### Relevant parts of the log file

```
-----------------------------------------------------------------------------------------------------
Build Error on Instance 'IdentityServer4.Services.ReturnUrlParser' 
    for PluginType IdentityServer4.Services.ReturnUrlParser
 - and Instance 'IdentityServer4.Services.DefaultIdentityServerInteractionService' for PluginType IdentityServer4.Services.IIdentityServerInteractionService

No default Instance is registered and cannot be automatically determined for type 'IdentityServer4.Stores.IAuthorizationParametersMessageStore'
There is no configuration specified for IdentityServer4.Stores.IAuthorizationParametersMessageStore


-----------------------------------------------------------------------------------------------------
Build Error on Instance 'IdentityServer4.Endpoints.AuthorizeCallbackEndpoint' 
    for PluginType IdentityServer4.Endpoints.AuthorizeCallbackEndpoint

No default Instance is registered and cannot be automatically determined for type 'IdentityServer4.Stores.IAuthorizationParametersMessageStore'
There is no configuration specified for IdentityServer4.Stores.IAuthorizationParametersMessageStore
```

brockallen:
If you've not configured this feature, then there's no code path that triggers the use of this service. I guess in Automapper it's mandatory, whereas in ASP.NET Core DI that is not.

brockallen:
We'll add a default.

evereq:
Resolved it in my project using (in Startup.cs / ConfigureServices method)
```
 services.AddAuthorizationParametersMessageStore<DistributedCacheAuthorizationParametersMessageStore>()
```
(see https://github.com/IdentityServer/IdentityServer4/blob/master/src/IdentityServer4/src/Stores/Default/DistributedCacheAuthorizationParametersMessageStore.cs)

Think dummy implementation can be easy created from https://github.com/IdentityServer/IdentityServer4/blob/master/src/IdentityServer4/src/Stores/Default/QueryStringAuthorizationParametersMessageStore.cs

[...]

PR #3894, opened 2019-12-10
Add default instance for IAuthorizationParametersMessageStore

brockallen:
#3552


leastprivilege:
Maybe we wait and make the query string a proper store in v4?

brockallen:
> Maybe we wait and make the query string a proper store in v4?

In 4, yes, we can do that.

PR #3895, opened 2019-12-10
use asynchronous EF methods

brockallen:
#2067

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -28,6 +28,11 @@ jobs:
       packageType: 'sdk'
       useGlobalJson: true
   - task: PowerShell@2
+    displayName: 'Start LocalDb'
+    inputs:
+      targetType: filePath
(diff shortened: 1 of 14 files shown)


AliBazzi:
Hi @brockallen ,

I think you overlooked [CorsPolicyService](https://github.com/IdentityServer/IdentityServer4/pull/3674/files#diff-d6171d6b788a414421fb4d52b1da4c9b).

And sorry for not making it apparent that a PR already exist for the same purpose.

brockallen:
Good catch. I have now added it to the cors service. Thanks.

leastprivilege:
The build seems to be dead-locking...

AliBazzi:
Same happened with my PR, I don't know why is that with windows build only

brockallen:
finally sorted out why it was deadlocking. done. merged. kthxbye.

Issue #3913, opened 2019-12-15
Why is acr_values not passed to POST /connect/token


natelaff:
Tried to ask this on SO but figured I need to go right to this source as maybe this is an enhancement request? Or maybe there is a better way to handle this than what I'm trying.

I have a multi-tenant app (DB per tenant) with a single IS4 instance using ASP.NET Core Identity and use IPersistedGrantDbContext to push persisted grants to each database.

When I attempt to authorize I pass acr_values tenant:tenantname and determine which database to connect to-- I inspect the HttpRequest and parse this out. This all works great... up until POST /connect/token (where presumably it is also reading/writing to PersistedGrants entity). Once this post happens, I lose track of my tenant name and everything fails. It's not included in the form data that's being posted.

I don't have an authorized user at this point so I don't have the tenant claim yet.

Is there a way to push acr_values to this final post? Is there a more obvious way to do this? Thanks!

brockallen:
It's from the OIDC spec and it's a flag parm to give hints on how to authenticate the user (e.f. MFA).

natelaff:
Hmm so is there a better way to pass information all the way from the start to finish? This the last call that I’m struggling with getting the necessary information of which db I’m trying to authenticate against in this multi tenant environment.

natelaff:
Ok, so setting a cookie doesn't work because of #3914 but I was able to get around the invalid grant error by using a separate db for grants. However, I still have the same issue essentially because /connect/token seems to be when ProfileService is invoked, so I still don't have the right database.

Any suggestions here? Surely I'm not the only person to have ever tried a user store in a database per tenant model. Can you pass the original request in the header or connect/token or something just so we can get to it for this purpose?

natelaff:
For the record I ended up going a completely different route and implementing IEndpointRouter to add a path segment. i.e. https://localhost:5000/tenantname which seems to work well. It requires a separate authority for each tenant, which is proving difficult to deal with in ASP.NET Core web app (can't set Authority dynamically based on context.) but so far this has proved to be the only solution where I can consistently get the token all the way through the flow.

Any other official thoughts on this would be helpful. In general some kind of multi-tenancy guidance would be good. I still don't understand why I couldn't get to that arc_values in the final /connect/token call. One small thing would have made this much easier.

brockallen:
acr_values is not a param from the spec for the token endpoint.

natelaff:
[...]

Issue #2595, opened 2018-09-07
External authentication failing on iOS 12


ryanmendoza:
Please only use the issue tracker for bug reports and/or feature requests. For general security questions, or free or commercial support options do __not__ use the issue tracker and instead see [here](https://identityserver4.readthedocs.io/en/release/intro/support.html) for more details.

For bug reports,  include the relevant log files related to your issue. See here how to enable [logging](https://identityserver4.readthedocs.io/en/release/topics/logging.html). Delete this line once you have.

Finally, please keep the issue concise and to the point. If you paste in more code than the text for the issue you are reporting then we will most likely not read it. 

### Issue / Steps to reproduce the problem

External authentication failing in iOS 12 Safari.

I am seeing failures across all iOS 12 (beta) devices when authenticating using an external provider. There is no issue with iOS <11 or mobile Firefox, Chrome (nor desktop/Android browsers).

` var result = await HttpContext.AuthenticateAsync(IdentityServer4.IdentityServerConstants.ExternalCookieAuthenticationScheme);` is failing for devices running Safari in iOS 12. Tracing the request, it looks like `idsrv.external` cookie is empty when redirecting from `signin-oidc` endpoint to the `externalcallback` endpoint, but is populated if you simply refresh the browser on the `externalcallback` page.

If you refresh the errored /externalcallback page, it will continue on and work fine.

I understand that iOS 12 is a beta OS, but it is very close to release and it would be nice to get this figured out.

**Update:** I was able to replicate this in Safari 12.0 for macOS Mojave as well. My guess is this may be more related to Microsoft authentication libraries...

### Relevant parts of the log file

```
2018-09-07 01:58:39.674 +00:00 [Error] Microsoft.AspNetCore.Server.Kestrel: Connection id "0HLGKCKUC9350", Request id "0HLGKCKUC9350:0000000C": An unhandled exception was thrown by the application.
System.Exception: External authentication error
   at IdentityServer.UI.AccountController.ExternalLoginCallback() in C:\Users\Ryan\Projects\Identity Server\src\IdentityServer\UI\Account\AccountController.cs:line 192
   at Microsoft.AspNetCore.Mvc.Internal.ActionMethodExecutor.TaskOfIActionResultExecutor.Execute(IActionResultTypeMapper mapper, ObjectMethodExecutor executor, Object controller, Object[] arguments)
[...]

PR #3914, opened 2019-12-15
Session Cookies need to be LAX

cboyce428:
Cookies are not being saved on a redirect, an easy way to see this is in postman.

To replicate problem get latest code and run this in the database first (IdentityServer4-master\samples\Quickstarts\4_EntityFramework)

INSERT [dbo].[ClientRedirectUris] ([RedirectUri], [ClientId]) VALUES (N'http://localhost:5002/', 1)
INSERT [dbo].[ClientGrantTypes] ([GrantType], [ClientId]) VALUES (N'authorization_code', 1)
INSERT [dbo].[ClientScopes] ([Scope], [ClientId]) VALUES (N'profile', 1)
INSERT [dbo].[ClientScopes] ([Scope], [ClientId]) VALUES (N'openid', 1)
update [Clients] set [AllowOfflineAccess] = 1 where [Id] = 1

Within postman
1) Create a new request
2) Click the Authorization tab
3) Change type to 'OAuth 2.0'
4) Click 'Get New Access Token'
5) Complete as follows
Call back URL: http://localhost:5002/
Auth URL: http://localhost:5000/connect/authorize
Access Token URL: http://localhost:5000/connect/token
Client ID: client
Client Secret: secret
Scope: api1 openid profile offline_access
State: <leave blank>

6) Then press 'Request Token'
7) Enter 'bob' & 'bob' without this change the page will simply reload and do nothing, if the cookies are changed it will work correctly.

Does not introduce breaking change as far as I can tell.

### DIFF
--- a/src/IdentityServer4/src/Configuration/DependencyInjection/ConfigureInternalCookieOptions.cs
+++ b/src/IdentityServer4/src/Configuration/DependencyInjection/ConfigureInternalCookieOptions.cs
@@ -31,7 +31,7 @@ namespace IdentityServer4.Configuration
                 options.ExpireTimeSpan = _idsrv.Authentication.CookieLifetime;
                 options.Cookie.Name = IdentityServerConstants.DefaultCookieAuthenticationScheme;
                 options.Cookie.IsEssential = true;
-                options.Cookie.SameSite = SameSiteMode.None;
+                options.Cookie.SameSite = SameSiteMode.Lax;
 
                 options.LoginPath = ExtractLocalUrl(_idsrv.UserInteraction.LoginUrl);
                 options.LogoutPath = ExtractLocalUrl(_idsrv.UserInteraction.LogoutUrl);
--- a/src/IdentityServer4/src/Services/Default/DefaultUserSession.cs
+++ b/src/IdentityServer4/src/Services/Default/DefaultUserSession.cs
@@ -260,7 +260,7 @@ namespace IdentityServer4.Services
                 Secure = secure,
                 Path = path,
                 IsEssential = true,
-                SameSite = SameSiteMode.None
+                SameSite = SameSiteMode.Lax
             };
 
             return options;


leastprivilege:
Postman is not a realistic test client. You might have heard about the incompatibilities wrt to SameSite and various browser.

This is really an ASP.NET Core issue - but we will provide some guidance in our specific context soon. @brockallen

brockallen:
I think we can close this - Lax is just not going to work.

cboyce428:
How about you simply remove 'SameSite' from the cookie instead?, I happen to use xero api which uses this project for their authentication (and they even have a mention on https://identityserver.io/).

One of their examples uses postman https://github.com/XeroAPI/xero-postman-oauth2 they must be using an old version or amended this as currently it doesn't work.

A previous change was made especially for IOS is postman really that bad?
 https://github.com/IdentityServer/IdentityServer4/issues/2595

rev23dev:
Is this why cookies I set aren't in the request for calls to things like:

- /connect/token
- /.well-known/openid-configuration
- /.well-known/openid-configuration/jwks

I am still fighting with https://github.com/IdentityServer/IdentityServer4/issues/3913 -- one proposed solution was to store tenant info in cookie and read back out, but yet again I've been foiled by not having cookies I've set available in these requests.

cboyce428:
Hi Rev23dev,

Yes this is exactly the problem I was having, I did a packet dump using fiddler & saw browser was not setting after the token redirect.. (same issue https too)

This seems like a browser issue search Google "302 redirect cookies samesite" but it's also broken in Edge so not specific to chrome weirdly lax cookies is supposed to be more restricted than none..

Try my tweak & please post an update otherwise going to just use my fork

rev23dev:
Hmm. I've been using in the Oidc Winforms client, which I guess would be IE. Broken there as well then.

PR #3926, opened 2019-12-19
IdentityErrorCode added to IdentityServer4.Constants

alexandra142:
IdentityErrorCode enum added to IdentityServer4.Constants

### DIFF
--- a/src/IdentityServer4/src/Constants.cs
+++ b/src/IdentityServer4/src/Constants.cs
@@ -4,6 +4,7 @@
 
 using IdentityModel;
 using IdentityServer4.Models;
+using Microsoft.AspNetCore.Identity;
 using System;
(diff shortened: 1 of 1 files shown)


brockallen:
What's the motivation for this? IdentityServer relies upon the hosting app for the user management system, and we don't require ASP.NET Identity.

alexandra142:
For me the motivation was to check the reason why IDS4 cannot create new user. 
If I was looking for existing solution I found 
https://stackoverflow.com/questions/19961648/how-to-localize-asp-net-identity-username-and-password-error-messagesSo I thought it could be useful.

leastprivilege:
As Brock said - this is not an IdentityServer concern - since it is not protocol related.

PR #3935, opened 2019-12-22
Fix user code param name in DeviceController

### LINKED ISSUE #3923: Wrong parameter name for user code in quickstart DeviceController
### Issue / Steps to reproduce the problem

- Make a device authorization request
- Open the `verification_uri_complete` in the browser
- The page ignores the user code in the query string and prompts the user to enter it

The problem is that the controller expects the parameter name to be named `user_code`, but it's actually named `userCode` in `verification_uri_complete`.

I realize that it's configurable, but the quickstart controller should match the default name, or use the name from the options. Maybe something like this:

```diff
         [HttpGet]
-        public async Task<IActionResult> Index([FromQuery(Name = "user_code")] string userCode)
+        public async Task<IActionResult> Index([FromServices] IOptions<IdentityServerOptions> options)
         {
+            string userCodeParamName = options.Value.UserInteraction.DeviceVerificationUserCodeParameter;
+            string userCode = Request.Query[userCodeParamName];
             if (string.IsNullOrWhiteSpace(userCode)) return View("UserCodeCapture");

             var vm = await BuildViewModelAsync(userCode);
```

### Relevant parts of the log file

N/A

thomaslevesque:

Fixes #3923 

No. The existing code didn't work as expected.

(There doesn't seem to be a place where this controller was tested, and adding test infrastructure for it would probably be out of scope for this PR)

### DIFF
--- a/samples/Quickstarts/2_InteractiveAspNetCore/src/IdentityServer/Quickstart/Device/DeviceController.cs
+++ b/samples/Quickstarts/2_InteractiveAspNetCore/src/IdentityServer/Quickstart/Device/DeviceController.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Linq;
 using System.Threading.Tasks;
+using IdentityServer4.Configuration;
 using IdentityServer4.Events;
 using IdentityServer4.Extensions;
 using IdentityServer4.Models;
@@ -13,6 +14,7 @@ using IdentityServer4.Stores;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 
 namespace IdentityServer.Device
 {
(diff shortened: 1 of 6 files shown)


samples/Quickstarts/2_InteractiveAspNetCore/src/IdentityServer/Quickstart/Device/DeviceController.cs
```diff
@@ -24,25 +26,30 @@ public class DeviceController : Controller
...
+            _options = options;
             _logger = logger;
         }
 
         [HttpGet]
-        public async Task<IActionResult> Index([FromQuery(Name = "user_code")] string userCode)
+        public async Task<IActionResult> Index()
         {
+            string userCodeParamName = _options.Value.UserInteraction.DeviceVerificationUserCodeParameter;
+            string userCode = Request.Query[userCodeParamName];
```
thomaslevesque: Since `VerificationUriComplete` is generated based on this option, I think it's the only way to properly reflect the configuration

leastprivilege:
@scottbrady91 please review

note to self: this must go in both v3 and v3.1.1

leastprivilege:
thanks!

PR #4015, opened 2020-01-24
Quickstart.UI native client detection heuristic based on client's redirect uri

### LINKED ISSUE #3997: Improve redirect treatment for clients in Quickstart.UI
_I suppose the idea here is covered by #3407, but this is more focused and constrained to just the Quickstart UI._

I'm wondering about the special treatment of PKCE clients in the QuickStart UI with regard to redirecting back to the client (e.g. [1]). This special redirection treatment assumes that every PKCE client is a native client. But I think this assumption is becoming less and less true. In newer AspNet Core versions, using PKCE is a matter of setting a couple of configuration properties and adding it to older AspNet Core is not much work either [2]. It's also possible with Katana 4.1 [3]. Also, we apparently need just two grants and one of them includes PKCE and covers all kinds of apps [4] ;)

So, I wonder if there is a better differentiator than whether RequirePkce is true. Would it make sense to base the decision on what the redirect uri is? If the scheme of the redirect uri is http or https, we do a regular http redirect. Otherwise, we show the intermediate screen. The assumption then is that a native client uses _Private-use URI Scheme Redirection_ [5].

[1] https://github.com/IdentityServer/IdentityServer4.Quickstart.UI/blob/6f187aa50f31690e44c838885b8f816c96fc417f/Quickstart/Account/AccountController.cs#L93
[2] https://www.scottbrady91.com/OpenID-Connect/ASPNET-Core-using-Proof-Key-for-Code-Exchange-PKCE
[3] https://www.scottbrady91.com/ASPNET/Refreshing-your-Legacy-ASPNET-IdentityServer-Client-Applications
[4] https://leastprivilege.com/2019/09/09/two-is-the-magic-number/
[5] https://tools.ietf.org/id/draft-ietf-oauth-native-apps-12.html#rfc.section.7.1

ArturDorochowicz:
Redirection to a native client uses a loading page instead of an http
redirect. Previously, determination of the type of the client was based on
whether it used PKCE. This PR changes it to be based on the redirect uri. If it
has http or https scheme, an http redirect is used. Otherwise, we assume it's
a native client and use a loading page. This is based on the fact that a
native client should use a private-use uri scheme.

Fixes #3997 

Not in IdentityServer itself.

**Other information**:

### DIFF
--- a/src/IdentityServer4/host/Quickstart/Account/AccountController.cs
+++ b/src/IdentityServer4/host/Quickstart/Account/AccountController.cs
@@ -90,9 +90,9 @@ namespace IdentityServer4.Quickstart.UI
                     await _interaction.GrantConsentAsync(context, ConsentResponse.Denied);
 
                     // we can trust model.ReturnUrl since GetAuthorizationContextAsync returned non-null
-                    if (await _clientStore.IsPkceClientAsync(context.ClientId))
+                    if (context.IsNativeClient())
                     {
-                        // if the client is PKCE then we assume it's native, so this change in how to
+                        // The client is native, so this change in how to
                         // return the response is for better UX for the end user.
                         //return View("Redirect", new RedirectViewModel { RedirectUrl = model.ReturnUrl });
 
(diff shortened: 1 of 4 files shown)


brockallen:
Thanks for this. Another recent PR broke this one, so we cherry picked the work you've done here. Thanks -- it's a nice improvement.

PR #4019, opened 2020-01-26
Add client setting to require request object

leastprivilege:


snothub:
@leastprivilege shouldn't there be an updated ef migration to add `RequireRequestObject` to db? After upgrading everything to the latest preview 4, this gives sql error on authorize request.

leastprivilege:
yes - I forgot it. Do you want to add it or @brockallen ?

PR #4024, opened 2020-01-28
Add JAR support

leastprivilege:
This PR changes the request object logic to be compatible with both OIDC ROs and JAR

### DIFF
--- a/src/IdentityServer4/src/Validation/Default/AuthorizeRequestValidator.cs
+++ b/src/IdentityServer4/src/Validation/Default/AuthorizeRequestValidator.cs
@@ -251,7 +251,8 @@ namespace IdentityServer4.Validation
                 
                 if (jwtRequestValidationResult.Payload.TryGetValue(OidcConstants.AuthorizeRequest.ClientId, out var payloadClientId))
                 {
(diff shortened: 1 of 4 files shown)


uringeller:
Have you decided on wether or not to handle the "very-special-case" of symmetrically encrypted ROs in a JAR request without OAuth 2.0 parameters (iow: no client_id)?

leastprivilege:
I support both client_id inside only - or in and outside the RO. But inside always takes precedence. And all inside values overwrite the outside values (if that makes sense).

steinarnoem:
That is good! But I was referring to the case where there are no OAuth parameters (outside the RO) and the token, for some unknown reason, is encrypted using a symmetric crypto algorithm. How would you find the correct key to decrypt?
I don't really think that it's a probable scenario, but maybe something that should be explicitly mentioned in the docs? Or just: "use the RAR endpoint to ensure confidentiality" :)

leastprivilege:
Well - this would not work. If it is symmetrically encrypted then the client_id needs to be outside the RO (but must match the one inside).

But that all hypothetical since we don't support encrypted ROs ;)

PR #4095, opened 2020-02-18
Return invalid_grant when redirect_uri is invalid on token endpoint

### LINKED ISSUE #4075: Error Response with invalid redirection URI
In the OAuth authorization_code flow, when validating the redirect_uri on the token endpoint, why do we return an `unauthorized_client` when the redirect_uri is invalid.
Shouldn't [this ](https://github.com/IdentityServer/IdentityServer4/blob/releases/3.1.x/src/IdentityServer4/src/Validation/Default/TokenRequestValidator.cs#L270) be an `invalid_grant` according to the spec:
> "invalid_grant
>                The provided authorization grant (e.g., authorization
>                code, resource owner credentials) or refresh token is
>                invalid, expired, revoked, does not match the redirection
>                URI used in the authorization request, or was issued to
>                another client."
> https://tools.ietf.org/html/rfc6749#section-5.2

Edit: mixed the authorize and token endpoint error-responses.

user1336:
Fixes #4075 

- [ X] The commit follows our [guidelines](https://github.com/IdentityServer/IdentityServer4/blob/master/.github/CONTRIBUTING.md)
- [ X] Unit Tests for the changes have been added (for bug fixes / features)


brockallen:
Thanks

PR #4140, opened 2020-03-04
Check for existing email claim

### LINKED ISSUE #4138: Email claim is added although there is already one
The UserClaimsFactory adds a email claim (and email_verified) although there is already one.

See https://github.com/IdentityServer/IdentityServer4/blob/3afba78dc11ebc73ba84a7b8142273e0d6e2caae/src/AspNetIdentity/src/UserClaimsFactory.cs#L55

As positive example, on adding the name claim, there is a check first.
https://github.com/IdentityServer/IdentityServer4/blob/3afba78dc11ebc73ba84a7b8142273e0d6e2caae/src/AspNetIdentity/src/UserClaimsFactory.cs#L45

twenzel:
Checks for an already existing "email" claim in the ClaimPrincipal before adding a new second.
This PR suppresses the duplicated email claim in the token.

Fixes #4138

### DIFF
--- a/src/AspNetIdentity/src/UserClaimsFactory.cs
+++ b/src/AspNetIdentity/src/UserClaimsFactory.cs
@@ -47,7 +47,7 @@ namespace IdentityServer4.AspNetIdentity
                 identity.AddClaim(new Claim(JwtClaimTypes.Name, username));
             }
 
-            if (_userManager.SupportsUserEmail)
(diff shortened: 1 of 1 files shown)


brockallen:
I had a look. Two extra things I'd ask:

* If there's already an email, can you compare the value and don't add it a second time if it's the same value.
* Can you do the same for the phone number claim?

Issue #4184, opened 2020-03-20
EndSession - Invalid ClientId when validating id_token_hint


ejmarunde:
I'm receiving the following error when calling the endSession endpoint with a valid id_token_hint:

  `System.InvalidOperationException: Client https://authorization-api-dev.XXXX.com/resources isn't configured`

I traced this to the default TokenValidator.**ValidateIdentityTokenAsync** making a request to **GetClientIdFromJwt**.  That method is returning the first **audience**, instead of the **client_id** claim.  The token validator is then calling FindEnabledClientByIdAsync and passing the tokens audience instead of client_id, which produces the above error.

Am I missing something here?  Is the first audience supposed to be the client_id?  I can't believe that I would be the only one with this issue.

I will submit a PR with the proposed fix to assist in reviewing this issue.

ejmarunde:
I can't seem to link the issue in the PR, but here's the PR.  https://github.com/IdentityServer/IdentityServer4/pull/4185

ejmarunde:
Never mind.  I figured it out.  I'm using Owin in a legacy MVC application and for some reason it's passing an access token instead of an id token.

PR #4185, opened 2020-03-20
Fixes token validation used by endsession endpoint for validating post logout redirect url

ejmarunde:
Fixes an error where TokenValidator.FindEnabledClientByIdAsync was returning the audience claim instead of client_id which was then used to lookup the current client.


ejmarunde:
I can't seem to link the issue to this PR, but here's the associated issue.  https://github.com/IdentityServer/IdentityServer4/issues/4184

ejmarunde:
Ahh...  I figured it out.  I'm using OWIN in a legacy MVC application and for some reason it's passing the access token instead of an id token.

PR #4226, opened 2020-03-29
Reduce usage of Newtonsoft.Json

leastprivilege:
I don't think we can remove it completely right now - but 80% in core is gone now.


src/IdentityServer4/src/Extensions/NameValueCollectionExtensions.cs
```diff
@@ -8,11 +8,32 @@
 using System.Collections.Specialized;
 using System.Text;
 using System.Text.Encodings.Web;
+using Microsoft.AspNetCore.SignalR.Protocol;
```
brockallen: plz fix this shit

src/IdentityServer4/test/IdentityServer.IntegrationTests/Clients/UserInfoClient.cs
```diff
@@ -66,7 +66,7 @@ public async Task Valid_client_with_GET_should_succeed()
             userInfo.Claims.Should().Contain(c => c.Type == "email_verified" && c.Value == "true");
         }
 
-        [Fact]
+        [Fact(Skip = "discuss")]
```
brockallen: pending

PR #4227, opened 2020-03-29
switch to named HTTP clients from factory (instead of typed)

leastprivilege:
Don't like the typed HTTP clients magic. Our new approach will

* list all names for our clients in `IdentityServerConstants.HttpClients`
* use those names internally when resolving our services from DI

If you want to customize a client, override the client in the factory using the name.


amadard:
I am curious, why don't you like the typed HttpClients?

leastprivilege:
I prefer the simplicity - and less DI magic.

Issue #3072, opened 2019-03-04
Front channel logout: empty post_logout_redirect_uri parameter still gets redirected.


alennartz:
### Issue 
In IdentityServer4 the selection of a post logout redirect uri for front channel logout has a fallback behavior for cases when the logout request does not specify a redirect uri. What it does is check if the client registration has exactly 1 post-logout redirect uri configured and if so, uses that in the validated request object.

see [EndSessionRequestValidator.cs](https://github.com/IdentityServer/IdentityServer4/blob/63a50d7838af25896fbf836ea4e4f37b5e179cd8/src/Validation/Default/EndSessionRequestValidator.cs)

specifically lines 143-156:
```
var redirectUri = parameters.Get(OidcConstants.EndSessionRequest.PostLogoutRedirectUri);
if (redirectUri.IsPresent())
{
    if (await UriValidator.IsPostLogoutRedirectUriValidAsync(redirectUri, validatedRequest.Client) == false)
    {
        return Invalid("Invalid post logout URI", validatedRequest);
    }

    validatedRequest.PostLogOutUri = redirectUri;
}
else if (validatedRequest.Client.PostLogoutRedirectUris.Count == 1)
{
    validatedRequest.PostLogOutUri = validatedRequest.Client.PostLogoutRedirectUris.First();
}
```

My preference would be to not redirect the user in this case but the only extension point i have at the moment to change this logic is to re-implement IEndSessionRequestValidator 

I was wondering if you would consider changing this behavior in IdentityServer4 directly or barring that, provide an easier way of overriding it. 

The specification for front channel logout does seem to imply that not redirecting in this situation is the desireable behavior. See [OIDC Session spec](https://openid.net/specs/openid-connect-session-1_0.html#RedirectionAfterLogout) or [OIDC Front channel logout spec](https://openid.net/specs/openid-connect-frontchannel-1_0.html#RPInitiated)
`In some cases, the RP will request that the End-User's User Agent to be redirected back to the RP after a logout has been performed. Post-logout redirection is only done when the logout is RP-initiated, in which case the redirection target is the post_logout_redirect_uri query parameter value used by the initiating RP; otherwise it is not done.`

However since the wording in the specification does not say that it "MUST NOT" be done I figured I'd open this point for discussion. 

Some options i see for ways to change IdentityServer4 

1.  Remove the else if portion with the fallback logic completely. (which may break existing implementations when they pull the new version, but brings the implementation closer to the spec.)
[...]

Issue #3470, opened 2019-07-24
IdentityServerConstants.ProfileDataCallers to have EndSessionRequest


Janidbest:
**Is your feature request related to a problem? Please describe.**

> This problem is a bit related to the old post https://github.com/IdentityServer/IdentityServer3/issues/1231 

When the users are requesting logout after something of the user critical information is changed( Ex. admin changed the username, so security stamp changed), is having issues to gracefully logout.

`EndSessionRequestValidator` is calling the `TokenValidator.ValidateIdentityTokenAsync` which is calling  `_profile.IsActiveAsync`. Since the user's security stamp is changed, the IsActive Async si returning "False". Regardless of the user active or inactive on logout requests, the flow should continue (without calling the profile.IsActive). 

As an alternative, we are showing the default IDM logout page if the PostLogoutRedirectUri is empty (which means IDM did not resolved the client from the request though id_token and the valid postlogoutredirecturl exist.

I suggest the following changes
 - Pass the context data caller as `EndSessionRequest` so the profile service `IsActiveContext.Caller `will give a provision to do a check and return "true". 

//profileService IsActiveAsync
` if (context.Caller == IdentityServerConstants.ProfileDataCallers.EndSessionEndPoint)
                context.IsActive = true; `

after all it make sense to clear the session and cookies of the user though they are not active or username or email address was changed.

Update after raising this request -

 I can see the `HttpContext.Request.Path `- /end/session, so for now we are using this check to set the `context.IsActive `to `true`

So this request for now is optional, and treat this as a low priority.

brockallen:
I'll have a look. The validator might be able to be more lax in some scenarios.

brockallen:
You can always change `IsActive` to use the caller to know if it's being checked from this scenario by comparing to the `IdentityServerConstants.ProfileIsActiveCallers.IdentityTokenValidation` constant value.

brockallen:
PR merged.

PR #4235, opened 2020-03-30
end session changes: IsActive no longer called and no longer default to a single redirect uri

brockallen:
#3072


brockallen:
While reviewing this, we should consider reviewing #3470 at the same time

PR #4243, opened 2020-04-01
Add reason to description so we know if it's a lockout.

Ben67366:
When an account is locked out, give the correct description.

No.


brockallen:
I don't know if I want this. Do you want to leak that info to the client so it can leak it to the end user?

Ben67366:
How about only leak the lockout information?
For users, if they forget their password, and want try their password's patterns, maybe that will help them.
And also, for hackers, it will let them know it will block them from bursting trying?\

PR #3978, opened 2020-01-11
StrictRedirectUriValidatorAppAuth allow path after port

VictorioBerra:
As per RFC8252 Section 7.3.  Loopback Interface Redirection here https://tools.ietf.org/html/rfc8252#section-7.3

**What issue does this PR address?**
https://github.com/IdentityServer/IdentityServer4/issues/3974

**Does this PR introduce a breaking change?**
No.

**Please check if the PR fulfills these requirements**
- [X] The commit follows our [guidelines](https://github.com/IdentityServer/IdentityServer4/blob/master/.github/CONTRIBUTING.md)
- [X] Unit Tests for the changes have been added (for bug fixes / features)

**Other information**:

Issue #3974, opened 2020-01-08
StrictRedirectUriValidatorAppAuth does not seem to work for mvc.code


VictorioBerra:
**We can only help you if you are on the latest version. Either use the latest 2.x or 3.x**

Please only use the issue tracker for bug reports and/or feature requests. For general security questions, or free or commercial support options do __not__ use the issue tracker and instead see [here](http://docs.identityserver.io/en/latest/intro/support.html) for more details.

For bug reports,  include the relevant log files related to your issue. See here how to enable [logging](https://identityserver4.readthedocs.io/en/release/topics/logging.html). Delete this line once you have.

Finally, please keep the issue concise and to the point. If you paste in more code than the text for the issue you are reporting then we will most likely not read it. 

### Issue / Steps to reproduce the problem
- Clone IS4 Repo
- build.ps1
- Start IS4 Host
- Edit mvc.code launchsettings to bind to `http://127.0.0.1:whatever` instead of `http://localhost:44392`
- Start mvc.code client
- Login fails.

### Relevant parts of the log file

```
[11:35:14 Information] Microsoft.AspNetCore.Authentication.Cookies.CookieAuthenticationHandler
idsrv was not authenticated. Failure message: Unprotect ticket failed

[11:35:14 Debug] IdentityServer4.Endpoints.AuthorizeEndpoint
No user present in authorize request

[11:35:14 Debug] IdentityServer4.Validation.AuthorizeRequestValidator
Start authorize request protocol validation

[11:35:14 Debug] IdentityServer4.Stores.ValidatingClientStore
client configuration validation for client mvc.code succeeded.

[11:35:15 Error] IdentityServer4.Validation.AuthorizeRequestValidator
Invalid redirect_uri: http://127.0.0.1:44392/signin-oidc
[...]

PR #4245, opened 2020-04-02
Strict redirect uri validator app auth with path

njparakh:
[#3974](https://github.com/IdentityServer/IdentityServer4/issues/3974)

No

- [&check; ] The commit follows our [guidelines](https://github.com/IdentityServer/IdentityServer4/blob/master/.github/CONTRIBUTING.md)
- [ &check;] Unit Tests for the changes have been added (for bug fixes / features)

**Other information**:
This PR is similar to [this one](https://github.com/IdentityServer/IdentityServer4/pull/3978) but does not use RegEx. It allows for a path in the Loopback address, but does not validate that path. 

I'd like to acknowledge [VictorioBerra](https://github.com/VictorioBerra) - the unit tests added in this PR are based on the unit tests he wrote in [his PR](https://github.com/IdentityServer/IdentityServer4/pull/3978).

### DIFF
--- a/src/IdentityServer4/src/Validation/Default/StrictRedirectUriValidatorAppAuth.cs
+++ b/src/IdentityServer4/src/Validation/Default/StrictRedirectUriValidatorAppAuth.cs
@@ -98,7 +98,7 @@ namespace IdentityServer4.Validation
             }
 
             string portAsString;
-            int indexOfPathSeparator = parts[2].IndexOfAny(new char[] { '/', '?' });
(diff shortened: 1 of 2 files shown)


brockallen:
Nice, thanks!

MichelJansson:
This is a great add.
But now with the SameSite cookie changes, everything needs to be HTTPS, even for loopback. Can we also allow https://127.0.0.1?

I realize it's only the http scheme that the [RFC](https://tools.ietf.org/html/rfc8252#section-7.3) covers - but would it be less safe to also allow https?

PR #4301, opened 2020-04-17
Update ExternalController.cs

Jay-study-nildana:
There is this major bug where, the scheme used was leading to a complete failure of external login. The thing is, the scheme is correctly used in previous quickstarts but was not updated here. 

So, applying that fix. now, with this update, the quickstart is fully functional for external login. Tested with EF with migrations and an online azure sql server database.

Fixes a not major but not minor issue where external login was just not working. now, it works with this fix.

it does not. 

**Other information**:

This simple fix will save hours for any other student who is learning this for the first time.


leastprivilege:
I think this scheme is correct since it is using ASP.NET Identity

PR #4303, opened 2020-04-19
update to new build script

adamralph:

Aligns the build script as much as possible with other repos, while keeping the multi-build approach as-is.

I have a few ideas about how to improve the multi-build thing, but they can come later.

No.

- ~Unit Tests for the changes have been added (for bug fixes / features)~

**Other information**:

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -5,21 +5,21 @@ New-Item -ItemType Directory -Force -Path ./nuget
 dotnet tool restore
 
 pushd ./src/Storage
-./build.ps1 $args
+Invoke-Expression "./build.ps1 $args"
 popd
 
 pushd ./src/IdentityServer4
-./build.ps1 $args
+Invoke-Expression "./build.ps1 $args"
 popd
 
 pushd ./src/EntityFramework.Storage
-./build.ps1 $args
+Invoke-Expression "./build.ps1 $args"
 popd
 
 pushd ./src/EntityFramework
-./build.ps1 $args
+Invoke-Expression "./build.ps1 $args"
 popd
 
 pushd ./src/AspNetIdentity
-./build.ps1 $args
+Invoke-Expression "./build.ps1 $args"
 popd
--- a/rebuild.ps1
+++ b/rebuild.ps1
@@ -1,2 +1,2 @@
 .\clean.ps1
-.\build.ps1 copy-pack-output
+.\build.ps1 quick
--- a/rebuild.ps1
+++ b/rebuild.ps1
@@ -1,2 +1,2 @@
 .\clean.ps1
-.\build.ps1 quick
+.\build.ps1 copy-pack-output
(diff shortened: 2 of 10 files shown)


build.ps1
```diff
@@ -5,21 +5,21 @@ New-Item -ItemType Directory -Force -Path ./nuget
 dotnet tool restore
 
 pushd ./src/Storage
-./build.ps1 $args
+Invoke-Expression "./build.ps1 $args"
```
adamralph, line 8: Without this, `$args` is passed as a single quote-wrapped value to `build.ps1`, so multiple options/targets cannot be passed.

src/AspNetIdentity/build/build.csproj
```diff
@@ -5,13 +5,12 @@
     <TargetFramework>netcoreapp3.0</TargetFramework>
   </PropertyGroup>
 
-  <ItemGroup>
-    <Compile Include="..\..\build\Program.Partial.cs" Link="Program.Partial.cs" />
+  <ItemGroup>
+    <Compile Include="..\..\build\Program.Partial.cs" Link="Program.Partial.cs" />
```
adamralph, line 9: The line endings were wrong on these two lines (CRLF instead of LF), and I can't figure out a way to get an editor to preserve them when making the change below.

This is part of a more general problem, in that `.editorconfig` has `end_of_line = crlf`, but almost every file in the repo has LF, which suggests that local `autocrlf` settings have been relied upon to control line endings. But this is a can of worms best opened another time...

rebuild.ps1
```diff
@@ -1,2 +1,2 @@
 .\clean.ps1
-.\build.ps1 quick
```
leastprivilege, line 2: You can't rename that - took me 3 months to teach @brockallen the quick.

He actually preferred to call it djquick

https://www.youtube.com/watch?v=J626Qcd7Glo

adamralph: But this is hidden inside `rebuild.ps1`. Do you also call it explicitly with `build.ps1 quick`? If you prefer, I can reinstate `quick`, simply pointing to `copy-pack-output` as a dependency.

leastprivilege: yes please ;)

adamralph: I pushed a fixup! commit to do that.

build.ps1
```diff
@@ -5,21 +5,21 @@ New-Item -ItemType Directory -Force -Path ./nuget
 dotnet tool restore
```
leastprivilege, line 5: This is not necessary anymore - right?

adamralph: It's still required, because I didn't put the `restore-tools` target into `Program.Partial.cs`. I figured that you probably wouldn't want that same thing running five times when you run the multi-build.

This is something I'm planning to address with my ideas around improving the multi-build. That said, I could put it into `Program.Partial.cs` now, to make each sub-build self-contained, but it would come at the cost of running that same target five times during the multi-build. It's your call.

leastprivilege: right - makes sense!

leastprivilege: Signing will never be done on an individual project - so it's fine the way it is.

leastprivilege:
thanks!

leastprivilege:
again.

PR #4335, opened 2020-04-29
Remove public origin setting

leastprivilege:
This setting was first introduced when people were struggling with forwarded header. 

Now we thing, setting up forwarded headers correctly is the right way to go. Public origin is a hack - and confusing.


leastprivilege:
docs!

PR #4341, opened 2020-04-30
Add Option to emit client ID as sub claim for non-interactive Flows

leastprivilege:
defaults to false


docmaster2:
src/Directory.Build.targets/brockallen/предлагается проверка лицензии.

PR #4352, opened 2020-05-01
Emit jti by default

leastprivilege:
to comply with JWT profile


slaneyrw:
Missed the default change in the XML doc comments for IncludeJwtId

Issue #3372, opened 2019-06-26
When Core 3 is released, use the new HostMatcherPolicy/HostAttribute in the diagnostic controllers for the quickstarts/samples.


VictorioBerra:
This is a placeholder issue for September-ish when Core 3 is released.

In Core 3 we are getting some new features that let us lock down our routes to only specific hosts, or ports https://github.com/aspnet/AspNetCore/pull/6214

This will be a perfect replacement for this code https://github.com/IdentityServer/IdentityServer4/blob/master/samples/Quickstarts/3_ImplicitFlowAuthentication/src/IdentityServer/Quickstart/Diagnostics/DiagnosticsController.cs#L19

VictorioBerra:
@leastprivilege example:

## Old

```csharp
    [SecurityHeaders]
    [Authorize]
    public class DiagnosticsController : Controller
    {
        public async Task<IActionResult> Index()
        {
            var localAddresses = new string[] { "127.0.0.1", "::1", HttpContext.Connection.LocalIpAddress.ToString() };
            if (!localAddresses.Contains(HttpContext.Connection.RemoteIpAddress.ToString()))
            {
                return NotFound();
            }

            var model = new DiagnosticsViewModel(await HttpContext.AuthenticateAsync());
            return View(model);
        }
    }
```

## New

```csharp
    [SecurityHeaders]
    [Authorize]
    [Host("127.0.0.1", "::1")]
    public class DiagnosticsController : Controller
    {
        public async Task<IActionResult> Index()
        {
            var model = new DiagnosticsViewModel(await HttpContext.AuthenticateAsync());
            return View(model);
        }
    }
```

Might want to triple check that `"::1"` is needed.

leastprivilege:
PR?

jcschmittmouts:
Can someone tell why only localAdresses can access diagnostics page?

VictorioBerra:
@jcschmittmouts because its full of a lot of sensitive information that you may not want to make available to a user or to an attacker.

PR #4356, opened 2020-05-01
feat(DiagnosticsController): Use `[Host("localhost:*")]`

VictorioBerra:
3.1#host-matching-in-routes-with-requirehost for more details.

https://github.com/IdentityServer/IdentityServer4/issues/3372

No

- [X ] The commit follows our [guidelines](https://github.com/IdentityServer/IdentityServer4/blob/master/.github/CONTRIBUTING.md)

**Other information**:
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-3.1#host-matching-in-routes-with-requirehost


brockallen:
I think the `Host` attribute indicates the target (by using the host request header), not filtering based on the user-agent. I don't think this is the same intent as the existing code is checking for.

VictorioBerra:
Ah I think you might be right.

PR #4382, opened 2020-05-10
Missing Scope

GioviQ:
I have troubles to make my IdentityServer setup working.
At the moment according to https://github.com/IdentityServer/IdentityServer4/blob/master/src/Storage/src/Models/ApiResource.cs
I discovered that api scope is not automatically created like stated in the current doc page.

Doc error


leastprivilege:
The resources config has changed for the upcoming v4. We will re-work the docs soon. thanks!

Issue #4431, opened 2020-05-24
Adding custom values in the filter for TokenRequestValidationLog


dajm316:
### Question
I am not sure if this can be considered a bug or a change request since I don't see a simple way to achieve what I am looking for however, here is the scenario:
In my solution, I am currently supporting custom grant types which may contain sensitive information in the request payload. If there is an error in validating the grant type, the framework, more specifically the **TokenRequestValidator**, will eventually call its **LogError** method. This eventually creates an instance of the _internal_ class **TokenRequestValidationLog** to generate the details that will forcibly append the details to the log. In the **TokenRequestValidationLog**  I see there is a sensitive value filter as follows:
```csharp
   private static readonly string[] SensitiveValuesFilter =
        {
            OidcConstants.TokenRequest.ClientSecret,
            OidcConstants.TokenRequest.Password,
            OidcConstants.TokenRequest.ClientAssertion,
            OidcConstants.TokenRequest.RefreshToken,
            OidcConstants.TokenRequest.DeviceCode

        };
```
 It would be great to make this extensible so that we can include other fields to redact in the logs.

To go back the original question. Is there an easy way to do this already? Or can this be flagged a bug or a change request?

leastprivilege:
Hey 

took a while ;)

But have a look

https://github.com/IdentityServer/IdentityServer4/pull/4444

Is that what you want?

PR #4444, opened 2020-05-27
Make sensitive data filters configurable

leastprivilege:
see #4431

### DIFF
--- a/src/IdentityServer4/src/Validation/Default/DeviceAuthorizationRequestValidator.cs
+++ b/src/IdentityServer4/src/Validation/Default/DeviceAuthorizationRequestValidator.cs
@@ -70,13 +70,13 @@ namespace IdentityServer4.Validation
 
         private void LogError(string message, ValidatedDeviceAuthorizationRequest request)
         {
-            var requestDetails = new DeviceAuthorizationRequestValidationLog(request);
+            var requestDetails = new DeviceAuthorizationRequestValidationLog(request, _options.Logging.TokenRequestSensitiveValuesFilter);
             _logger.LogError(message + "\n{requestDetails}", requestDetails);
         }
 
         private void LogError(string message, string detail, ValidatedDeviceAuthorizationRequest request)
         {
-            var requestDetails = new DeviceAuthorizationRequestValidationLog(request);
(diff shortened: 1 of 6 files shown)


src/IdentityServer4/src/Validation/Default/DeviceAuthorizationRequestValidator.cs
```diff
@@ -70,13 +70,13 @@ private DeviceAuthorizationRequestValidationResult Invalid(ValidatedDeviceAuthor
 
         private void LogError(string message, ValidatedDeviceAuthorizationRequest request)
         {
-            var requestDetails = new DeviceAuthorizationRequestValidationLog(request);
+            var requestDetails = new DeviceAuthorizationRequestValidationLog(request, _options.Logging.TokenRequestSensitiveValuesFilter);
```
dajm316, line 73: The name of the property is TokenRequestSensitiveValuesFilter but it's being used in the DeviceAuthorizationRequestValidationLog. Could this be confusing for someone?

leastprivilege: Isn't that in essence also a token request? Do you have a better idea?

dajm316: Hmm, the way I understood device authorization endpoint was to authorize a client application and to get back as a result, a user code, a device code, and verification url (and other fields) that a user can then use in a browser to authenticate themselves while the client application pokes the token endpoint.
So my perception is it's not really a token request from what I understand. It's really a device authorization request.

An alternative would be to add another property in the LoggingOptions named DeviceAuthorizationRequestSensitiveValuesFilter with the clientId and client secret.

And yet another alternative, which I think might be the best, is to leave the DeviceAuthorizationRequestValidationLog as is (i.e. remove the changes that were done) and keep the fixed list of sensitive values defined in the class. This request is not typically open for customization like the token request is.

leastprivilege: Agreed. thanks!

Issue #4492, opened 2020-06-09
Infinite loop in Token Cleanup after concurrency exception


ArturDorochowicz:
### Issue / Steps to reproduce the problem

There is a bug in `TokenCleanupService` which makes it go into an infinite loop (nearly certainly) after `DbConcurrencyException`.

When concurrency exception is caught, the loop continues without resolving the conflict in the `DbContext`. This means that on subsequent iteration, `DbContext` will want to delete a union of the grants from the previous iteration(s) and the newly selected ones (depending on the traffic in the database, the newly selected grants may mostly be the same as previously selected). The crucial thing is that the deletion of the grant(s) that caused the conflict will be attempted again and will inevitably fail the operation again. 

This results in executing and rolling back transactions in a tight loop. This may exhaust disk space for transaction log which actually can be a good thing, because it causes an exception that finally breaks the loop .

The same issue is repeated in `RemoveGrantsAsync` as well as `RemoveDeviceCodesAsync`.

Using IdentityServer4 3.1.3

https://github.com/IdentityServer/IdentityServer4/blob/92ef8cc35f9b56d0f6e62d4ab9390c8cdb892a4f/src/EntityFramework.Storage/src/TokenCleanup/TokenCleanupService.cs#L71-L142

### Relevant parts of the log file

```
<log goes here>
```

ArturDorochowicz:
@brockallen There's one more, unrelated thing about TokenCleanupService that I've learnt when investigating the issue.

This code: `.Where(x => x.Expiration < DateTime.UtcNow)` is translated to `where Expiration < getutcdate()`, at least when using SqlServer. This means that the expiration here is based on the database server's time whereas everywhere else in IdentityServer it's based on application server's time.

brockallen:
I think we use UTC everywhere, no?

ArturDorochowicz:
The thing is that it's time on the database server vs time on the application server. With sharp cut off (`< Now` as opposed to let's say `< Now - 1day`) and not synchronized clocks, the cleanup may very well be deleting grants that the application would still consider valid.

ArturDorochowicz:
I think a simple code change would make it use application server's time and not db server's time:
```csharp
var threshold = DateTime.UtcNow; // or DateTime.UtcNow - someTimespan;
...
.Where(x => ex.Expiration < threshold)
...
```

PR #4498, opened 2020-06-10
fix infinite loop in Token Cleanup after concurrency exception 

brockallen:
#4492


brockallen:
> One thing to note is that you don't inspect the type of the conflicted entry and make an assumption that these are all persisted grants. 

Yea, as I was working on it I did have that type of check and realized given the nature of what this was being used for I felt it was unnecessary.

ArturDorochowicz:
@brockallen Any chance for 3.1 release with this? :crossed_fingers:

aadreja:
> #4492

@brockallen I am using 4.0.2 and facing same error. Can you help

PR #4502, opened 2020-06-11
Update form content check to reject multipart forms

leastprivilege:


VictorioBerra:
This is a breaking change.

PR #4530, opened 2020-06-17
Reset `ConsumedTime` when saving soon-to-be-new refresh token

jakubfijalkowski:
None.

No.

**Other information**:
When using one-time refresh tokens, the code re-uses the `RefreshToken` instance to create new token. Unfortunately, this means that it also re-uses the `ConsumedTime` property. It isn't saved to the database directly (there will be `null` in the column), but it will get serialized to the `Data` JSON that is then used to detect if the token has been consumed.
This effectively block token refreshes after the first refresh.

In this PR I just reset the `ConsumedTime` back to null before storing the new token.


brockallen:
Great catch and thanks for the unit test! :)

PR #4575, opened 2020-06-25
Reduce log level for expired secrets

mpomery:
There is value in seeing if a client has expired secrets when attempting to validate as it helps debug issues, but I do not think it should be a warning.
Secrets validator gets called a lot so if you watch your logs for warnings you see a lot of them if you have a client with an expired secret.

For our implementation that uses IdentityServer4, we watch out logs for warning messages.
When a client has a secret expire we start to see a lot of warnings (triggering emails to tell us about them).
This PR changes the log level for expired secrets to info as we think it makes more sense.

No

**Other information**:


leastprivilege:
thanks!

Issue #4572, opened 2020-06-24
Exception with prompt=login in 4.0


crazyfx1:
### Issue / Steps to reproduce the problem
After upgrading to 4.0, an authorization request with prompt=login throws an exception.

```
https://xxx/connect/authorize?client_id=config&redirect_uri=xxx&response_type=code&scope=openid%20profile%20email&state=xxx&code_challenge=xxx&code_challenge_method=S256&prompt=login&response_mode=query
```

I'd assume this was introduced by #4208 

### Relevant parts of the log file

```
System.ArgumentNullException: Value cannot be null. (Parameter 'source')
   at System.Linq.ThrowHelper.ThrowArgumentNullException(ExceptionArgument argument)
   at System.Linq.Enumerable.Contains[TSource](IEnumerable`1 source, TSource value, IEqualityComparer`1 comparer)
   at System.Linq.Enumerable.Contains[TSource](IEnumerable`1 source, TSource value)
   at IdentityServer4.ResponseHandling.AuthorizeInteractionResponseGenerator.ProcessInteractionAsync(ValidatedAuthorizeRequest request, ConsentResponse consent)
   at IdentityServer4.Endpoints.AuthorizeEndpointBase.ProcessAuthorizeRequestAsync(NameValueCollection parameters, ClaimsPrincipal user, ConsentResponse consent)
   at IdentityServer4.Endpoints.AuthorizeEndpoint.ProcessAsync(HttpContext context)
   at IdentityServer4.Hosting.IdentityServerMiddleware.Invoke(HttpContext context, IEndpointRouter router, IUserSession session, IEventService events)
   at IdentityServer4.Hosting.IdentityServerMiddleware.Invoke(HttpContext context, IEndpointRouter router, IUserSession session, IEventService events)
   at IdentityServer4.Hosting.MutualTlsEndpointMiddleware.Invoke(HttpContext context, IAuthenticationSchemeProvider schemes)
   at Microsoft.AspNetCore.Authentication.AuthenticationMiddleware.Invoke(HttpContext context)
   at IdentityServer4.Hosting.BaseUrlMiddleware.Invoke(HttpContext context)
```

brockallen:
Thanks for finding that. PR submitted to fix.

PR #4577, opened 2020-06-25
fix exception with prompt=login

brockallen:
fixes bug/regression with the new prompt collection feature in 4.0

#4572


sowa111111:
@brockallen @leastprivilege Any chance to release that soon?

PR #4649, opened 2020-07-13
Fix 401 malformed WWW-Authenticate

mouxhun:
references
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate
https://tools.ietf.org/html/rfc7235#section-4.1

401 malformed WWW-Authenticate cause some error when java invoke api in IdentityServer4

No

**Other information**:


brockallen:
Not sure the value "Restricted" makes sense. Perhaps use "IdentityServer".

mouxhun:
Saw "Restricted" on nginx, i think both are ok, now it has been changed to IdentityServer

PR #4695, opened 2020-07-29
updating code

Jay-study-nildana:
I think the class format has changed. the updated code seems to the right one.

I think the class has changed and the doc is not updated.

no, it does not. 

**Other information**:

### DIFF
--- a/docs/topics/resources.rst
+++ b/docs/topics/resources.rst
@@ -52,8 +52,8 @@ The following example shows a custom identity resource called *profile* that rep
         {
             new IdentityResource(
                 name: "profile",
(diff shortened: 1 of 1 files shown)


bryantlikes:
@Jay-study-nildana looks like you only need to rename `claimTypes` to `userClaims` for it to work. That would keep with the pattern of passing the parameter names. 

``` c#
new IdentityResource(
                name: "profile",
                userClaims: new[] { "name", "email", "website" },
                displayName: "Your profile data")
```

leastprivilege:
Thanks. I think this has been fixed in another PR.

PR #4696, opened 2020-07-29
Correcting documentation issue with Windows Authentication

HuntJason:
IdentityServerConstants.ExternalCookieAuthenticationScheme ("idsrv.external") causes failure in the Callback action whose first line calls HttpContext.AuthencicateAsync(IdentityConstants.ExternalScheme);

Changing to 
await HttpContext.SignInAsync(
                IdentityConstants.ExternalScheme,

allows it to succeed.


brockallen:
This line depends on if you're using ASP.NET Identity or not, so it's not always what you're proposing.

PR #4703, opened 2020-07-31
Change from Encoding.UTF8.GetBytes to Convert.FromBase64String

vasicvuk:
Change from Encoding.UTF8.GetBytes to Convert.FromBase64String as Encoding.UTF8.GetBytes causes following error on Linux:
Interop+Crypto+OpenSslCryptographicException: error:0D07803A:asn1 encoding routines:ASN1_ITEM_EX_D2I:nested asn1

Issue with wrong sample for application working on both Windows and Linux

No

**Other information**:

### DIFF
--- a/docs/topics/mtls.rst
+++ b/docs/topics/mtls.rst
@@ -43,7 +43,7 @@ If you are using Nginx (which we found is the most flexible hosting option), you
 
             if(!string.IsNullOrWhiteSpace(headerValue))
             {
-                var bytes = Encoding.UTF8.GetBytes(Uri.UnescapeDataString(headerValue));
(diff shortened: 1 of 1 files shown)


brockallen:
No response from OP

vasicvuk:
@leastprivilage What exception?

leastprivilege:
I can't remember - probably because it is not base64 encoded.

What you are proposing cannot work - because the 2 APIs do fundamentally different things.

vasicvuk:
Nginx puts Base64 encoded raw certificate (DEX) when forwarding to header that is why Convert.FromBase64String is used

leastprivilege:
Show me your nginx config for the header forwarding

PR #4712, opened 2020-08-05
fix multiple WWW-Authenticate header to one

mouxhun:
WWW-Authenticate header should only have one

every WWW-Authenticate header should atleast have one auth schema, because we only used bearer token authentication, we should only have one WWW-Authenticate header.Currently it returns two likes below

> WWW-Authenticate: Bearer
> WWW-Authenticate: error="invalid_token"

test on https://demo.identityserver.io/connect/userinfo

No

**Other information**:


mouxhun:
reference https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.47

> The field value consists of at least one challenge that indicates the authentication scheme(s) and parameters applicable to the Request-URI

the second one don't have schema

brockallen (approved):

PR #4744, opened 2020-08-12
Introduce LoggingOptions.AuthorizeRequestSensitiveValuesFilter

mustakimali:
We take some sensitive parameters with the "AuthorizeRequest" endpoint with some ICustomAuthorizeRequestValidator implementation to validate them.

We don't want those sensitive parameters be logged when the validation fails.

This commit introduces `AuthorizeRequestSensitiveValuesFilter` similar to `TokenRequestSensitiveValuesFilter`. This will allow us customize which parameter gets scrubbed.


leastprivilege:
Thanks!

PR #4818, opened 2020-09-02
Update 5_entityframework.rst

Abrahamberg:
In the is4ef template there is no ApiResources, therefore it causes the error below and it will be impossible to follow the tutorial :
'Config' does not contain a definition for 'Apis'

Since ApiScopes exists in that template it should be replaces by   ApiResources


brockallen:
I think we got this in another PR already Thx!

PR #4839, opened 2020-09-08

FaridAhamat:
When defining new Client, we need to assign array type since AllowedGrantTypes is a type of Collection.

The code example is wrong (or outdated? I have never worked with previous version of IS so I wouldn't know). The AllowedGrantTypes is a Collection so we can't just assign a simple variable.

No

**Other information**:

### DIFF
--- a/docs/quickstarts/1_client_credentials.rst
+++ b/docs/quickstarts/1_client_credentials.rst
@@ -89,7 +89,7 @@ For this, add a client definition::
                 ClientId = "client",
 
                 // no interactive user, use the clientid/secret for authentication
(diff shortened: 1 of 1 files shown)


leastprivilege:
I think this code works fine. Are you saying it doesn't ?

FaridAhamat:
I'm gonna close this one. Now I know why it doesn't work: The Intellisense didn't give me "GrantTypes"! I thought the documentation is wrong, but it seems not. I only noticed this after Intellisense broke on several other types that I'm already familiar with... and then I realized it's the Intellisense that is broken and not the documentation

Issue #4810, opened 2020-09-01
Identity Server and single file published apps


javiercn:
Hi @leastprivilege and @brockallen.

We've recently updated the integration in ASP.NET Core to use Identity Server 4.0.4. It was very smooth! so great job there ❤️ 

That said, we found a bug though our manual validation process that relates to publishing single file applications. I found the root cause to be the access to the assembly location here which does not exist for single file published apps https://github.com/IdentityServer/IdentityServer4/blob/main/src/IdentityServer4/src/Configuration/IdentityServerApplicationBuilderExtensions.cs#L57

The impact is that this blocks ASP.NET Core apps that use the latest identity server from being published as a single file.

I'm not sure how problematic this is to fix, could an alternative be to stamp the version you are looking for in an assembly attribute and use that instead to retrieve the value? Would it be ok to omit the version if the app is single file publish?

Thanks in advance, and If there's anything we can do to help let us know

leastprivilege:
Hey, 

do you know a way to retrieve that value using a compatible API? If not possible, the short term solution would be to disable the version logging for those incompatible scenarios.

Either way - please create a PR.

thanks!

leastprivilege:
@javiercn any update?

javiercn:
@leastprivilege thanks for the response.

I'm wrapping up some other tasks, but I'll hope to get to this later this week. I'll have to look a bit more in depth to understand what the value is and what is used for. Your solution of disabling the version login in this case sounds good to me.

leastprivilege:
Well - if there is a way to get the same value from your new fancy single file scenario - that would be ideal ;)

leastprivilege:
merged

PR #4868, opened 2020-09-10
Make identity server work with publish single file in .NET 5.0

javiercn:
https://github.com/IdentityServer/IdentityServer4/issues/4810

No

I couldn't find unit tests in the area testing the current behavior, I'm happy to add one if you point me where to do so.

**Other information**:
Manual validation of the changes.
image


leastprivilege:
thanks!

brockallen:
Was this change backwards compat w/ 3.1?

javiercn:
@brockallen it should, but this is not an issue in 3.1 since the publish single file strategy is different (the assemblies we packaged and unzipped on a temp directory, while now they are mapped directly into memory, so there's no physical file at all, which was the root cause of the issue).

In general, is not "safe" to use `Assembly.Location` AFAIK if you want to support publish single file, we had to do work in MVC too to avoid issues there too.

PR #4978, opened 2020-10-16
Using a constant instead of string magic

igor-moskvitin:
**IdentityServer4 is feature complete. All new feature work is happening in our new [organization](https://github.com/duendesoftware). This means we can only accept bug fix PRs.**

small fix in documentation

no


leastprivilege:
Thanks - I prefer the strings to actually see what you get....

igor-moskvitin:
> Thanks - I prefer the strings to actually see what you get....

it makes sense, but on the other hand, modern IDEs deal with it as well

PR #5047, opened 2020-11-23
remove redundant secret description access

YVEF:
A little refactoring, to prevent unnecessary checking and access to a secret description


leastprivilege:
Good catch. thanks!

PR #5106, opened 2021-01-01
RefreshToken store fix

UsmanSabir:

Upon "grant_type" of "refresh_token", newly generated refresh token was being save with OLD access token. This PR fix this issue and now it will store new access token with new refresh token

No

- [ Yes] The commit follows our [guidelines](https://github.com/IdentityServer/IdentityServer4/blob/main/.github/CONTRIBUTING.md)
-
Verified
**Other information**:

### DIFF
--- a/src/IdentityServer4/src/ResponseHandling/Default/TokenResponseGenerator.cs
+++ b/src/IdentityServer4/src/ResponseHandling/Default/TokenResponseGenerator.cs
@@ -204,6 +204,7 @@ namespace IdentityServer4.ResponseHandling
 
             var oldAccessToken = request.ValidatedRequest.RefreshToken.AccessToken;
             string accessTokenString;
+            Models.Token newAccessToken = null;
 
             if (request.ValidatedRequest.Client.UpdateAccessTokenClaimsOnRefresh)
             {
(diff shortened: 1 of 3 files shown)


leastprivilege:
Could you explain the problem a bit more detailed?

UsmanSabir:
I was writing custom persistence for IdentityServer4. During unit tests, I faced an issue
1- Generated Access Token(reference type) + Refresh Token. 
2- Used jti and a custom field to relate(CorrelationId) access token with its refresh token. This is to track tokens chain
3- When Access token got expired, generated new access token from refresh token flow
4- Now I received both Access Token and the new Refresh token
5- Now here is the problem, the newly generated refresh token **was being saved with older/expired access token**. Ideally, it **should "persist" new refresh token with new Access token**, generated in step 4. 

To fix this issue, I am passing new Access Token as parameter to refresh token service so it can update and persist the correct pair of tokens. Also sets its default value in paramter so it doesn't break other flows.

malylemire1:
Had the same problem when using JwtId as cache identifier for a blacklist of access tokens.

malylemire1 (approved):

malylemire1 (approved):

PR #5121, opened 2021-01-12
Specify the default for RefreshTokenExpiration in the documentation

### LINKED ISSUE #5120: The default for RefreshTokenExpiration is missing in the documentation
We're working on changing the `RefreshTokenExpiration` from `Absolute` to `Sliding`, and noticed that the default is not specified in either of the 2 places that this is documented (the [Refresh Tokens](https://identityserver4.readthedocs.io/en/latest/topics/refresh_tokens.html#additional-client-settings), and [Client](https://identityserver4.readthedocs.io/en/latest/reference/client.html?highlight=RefreshTokenExpiration#token)), where most other settings have a default.

Is there a reason that this is left out, or can I put in a PR to add it?

tiesmaster:
Fixes #5120

### DIFF
--- a/docs/reference/client.rst
+++ b/docs/reference/client.rst
@@ -75,7 +75,7 @@ Token
     
     ``OneTime`` the refresh token handle will be updated when refreshing tokens. This is the default.
 ``RefreshTokenExpiration``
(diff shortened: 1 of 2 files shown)


leastprivilege:
thanks!

Issue #5134, opened 2021-01-21
Wrong Token error when using password flow with disabled user


reikla:
When a user is disabled via IProfileServices.IsActive() and logging in via ResourceOwnerCredentials, Identityserver returns invalid_grant. I think access_denied would be more appropriate. 

Imo the problem is to be found here: https://github.com/IdentityServer/IdentityServer4/blob/45f8cb5d2215263528b6481ad9571729d7090c96/src/IdentityServer4/src/Validation/Default/TokenRequestValidator.cs#L338

### Issue / Steps to reproduce the problem
Log in with a disabled user with password flow then you get "invalid_grant" as response

I will create a PR to fix that issue.

BR Reimar

PR #5136, opened 2021-01-21
Correct the response for a token when a user signs in with in an inac…

reikla:

#5134 When a user is disabled via IProfileServices.IsActive() and logging in via ResourceOwnerCredentials, Identityserver returns invalid_grant.

No

**Other information**:


leastprivilege:
The grant, which is the username/password, is invalid. What's wrong with that?

reikla:
I thought the grant username/password is valid in general, but the acces is denied due to a disabled user account.

PR #5271, opened 2021-06-04
Update refresh_tokens.rst

whatisthejava:
Been caught with this twice now.

If you leave the refresh token empty or use an old refresh_token an invalid_grant error is returned.  This is not helpful as I spent ages double checking everything about the request. 

I think a simple note is a good stop gap, because a more accurate fix would require a new error_type to be added which is possibly not wanted at this point 

(If a bigger fix is required, respond to PR and I will try and fix it )

Using an old refresh token gives an invalid_grant error, presuming this is correct behaviour,  the reader needs to be informed that the refresh_token must be valid

Inaccurate documentation 

No 

**Other information**:


leastprivilege:
Fair enough. Thanks!

